PR #30, opened 2017-11-30
Get Account API: Get username/email from ID

shashankmehra:
Update documentation

### DIFF
--- a/api/accounts/get_account_test.go
+++ b/api/accounts/get_account_test.go
@@ -1,0 +1,43 @@
+package accounts_test
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/keratin/authn-server/api/accounts"
+	"github.com/keratin/authn-server/api/test"
+	"github.com/keratin/authn-server/lib/route"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetAccount(t *testing.T) {
+	app := test.App()
+	server := test.Server(app, accounts.Routes(app))
+	defer server.Close()
+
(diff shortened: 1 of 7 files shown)


api/accounts/get_account_test.go
```diff
@@ -0,0 +1,43 @@
...
+	})
+
+	t.Run("valid account", func(t *testing.T) {
+		account, err := app.AccountStore.Create("unlocked@test.com", []byte("bar"))
+		require.NoError(t, err)
+
+		res, err := client.Get(fmt.Sprintf("/accounts/%v", account.ID))
+		require.NoError(t, err)
+		assert.Equal(t, http.StatusOK, res.StatusCode)
+		test.AssertGetAccountResponse(t, res, account)
```
cainlevy: i don't anticipate this being reused. would you please make it a private func in this file?

docs/guide-implementing_signup.md
```diff
@@ -32,6 +32,11 @@ process that submits with two API calls.
 > the data client-side (step 3) is important because it will significantly reduce the odds of this
 > happening.
 
+> NOTE:
```
cainlevy: 💯 for adding these docs. thanks!

shashankmehra:
@cainlevy Made the changes.

shashankmehra:
Weird. Its passing locally but it should not. Fixing.

shashankmehra:
@cainlevy The function `AssertGetAccountResponse` uses `extractResult` which is part of `asserts.go`. To move `AssertGetAccountResponse` would require to copy that function there too. I think to avoid code duplication we should let it be. Even though that function will not be reused, it would make sense to have it there from code organisation point of view. Let me know if you still want to move it.

cainlevy:
thanks for the pr! 🎉

Issue #40, opened 2018-01-28
error during make vendor


sgraham785:
while following the instructions in README I encountered a few issues while trying to get the dev environment setup. I will open this issue to track the fixes in my pull request
- Mac OSX High Sierra
- go version go1.9.3 darwin/amd64

these issues were encountered while running `make vendor` (Step 2):
```
go install
# github.com/keratin/authn-server/lib/compat
../../keratin/authn-server/lib/compat/key_id.go:20:9: undefined: jose.JSONWebKey
make: *** [vendor] Error 2
```
this issue is due to the import of github.com/square/go-jose, which should be gopkg.in/square/go-jose.v2 per the [docs](https://github.com/square/go-jose#versions)

once that is cleared up we get this error: (ran `make clean && make vendor`)
```
go install
# github.com/keratin/authn-server/api/meta
api/meta/get_root.go:14:3: undefined: views.Root
make: *** [vendor] Error 2
```
which is because `/api/views/root.ego` has not been compiled, this leads to a more abstract issue, where the dep cmd ego is not installed in the `Makefile` (although it is in `travis.yml` and `Dockerfile.builder`), so if you have never used ego then it won't be installed and the thus the error.

Adding it to `make vendor` fixes this but sort of "breaks" the `Makefile` workflow or could add bloat, so I changed `vendor -> init` and have it call `$(EGOS) 

I would also suggest adding `ego` to `glide.yaml` as a dep, but since you left it out I will also at this time

Let me know your thoughts?

PR #41, opened 2018-01-28
Fix dev env setup

sgraham785:
#40

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,6 @@ before_install:
   - sudo add-apt-repository ppa:masterminds/glide -y
   - sudo apt-get update -q
   - sudo apt-get install glide -y
-  - go get github.com/benbjohnson/ego/cmd/ego
 script: make test-ci
 
 after_success:
--- a/Makefile
+++ b/Makefile
@@ -7,17 +7,18 @@ MAIN := main.go routing.go
 
 .PHONY: clean
 clean:
-	rm -rf vendor
+	rm -rf init
 	rm -rf dist
 	rm -f api/views/*.ego.go
 
 # Process .ego templates (skippable)
 EGOS := $(shell find . -name *.ego | sed -e s/.ego/.ego.go/)
 $(EGOS):
+	go get github.com/benbjohnson/ego/cmd/ego
 	ego api/views
 
 # Fetch dependencies
-vendor: glide.yaml
+init: $(EGOS)
 	glide install
 	go install
 
@@ -29,7 +30,7 @@ linux-builder:
 
 # The Linux target is built using a special Docker image, because this Makefile assumes the host
 # machine is running MacOS.
-dist/linux/amd64/$(PROJECT): $(EGOS) vendor
+dist/linux/amd64/$(PROJECT): init
 	make linux-builder
 	docker run --rm \
 		-v $(PWD):/go/src/github.com/$(NAME) \
@@ -41,7 +42,7 @@ dist/linux/amd64/$(PROJECT): $(EGOS) vendor
 	bzip2 -c "$@" > dist/authn-linux64.bz2
 
 # The Darwin target is built using the host machine, which this Makefile assumes is running MacOS.
-dist/darwin/amd64/$(PROJECT): $(EGOS) vendor
+dist/darwin/amd64/$(PROJECT): init
 	GOOS=darwin GOARCH=amd64 CGO_ENABLED=1 go build -ldflags "-X main.VERSION=$(VERSION)" -o "$@"
 	bzip2 -c "$@" > dist/authn-macos64.bz2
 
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -14,10 +14,11 @@ clean:
 # Process .ego templates (skippable)
 EGOS := $(shell find . -name *.ego | sed -e s/.ego/.ego.go/)
 $(EGOS):
+	go get github.com/benbjohnson/ego/cmd/ego
 	ego api/views
 
 # Fetch dependencies
-vendor: glide.yaml
+init: $(EGOS)
```
cainlevy: i'm unsure about this change. would you make the `init` rule separate? my thinking was that `make vendor` should only run if `glide.yaml` has been recently modified. (https://www.gnu.org/software/make/manual/make.html#Prerequisite-Types)

Makefile
```diff
@@ -29,7 +30,7 @@ linux-builder:
 
 # The Linux target is built using a special Docker image, because this Makefile assumes the host
 # machine is running MacOS.
-dist/linux/amd64/$(PROJECT): $(EGOS) vendor
+dist/linux/amd64/$(PROJECT): init
```
cainlevy: i think replacing `$(EGOS) vendor` with `init` here and below will defeat build caching for these steps.

maybe that's okay? the caching doesn't pay attention to changes in source files, so executing this step each time might be safest.

Makefile
```diff
@@ -14,10 +14,11 @@ clean:
 # Process .ego templates (skippable)
 EGOS := $(shell find . -name *.ego | sed -e s/.ego/.ego.go/)
 $(EGOS):
+	go get github.com/benbjohnson/ego/cmd/ego
```
cainlevy: 👍

Makefile
sgraham785: I don't think that logic applies here. If it was back the way it was `vendor: glide.yaml` and run `make vendor` then simply run `make vendor` again you will see that it still runs `glide install` and so on, even though `glide.yaml` is unchanged.  

What you are actually doing is making that target dependent on `glide.yaml` as a prerequisite. So if `glide.yaml` is not present you get this error: `make: *** No rule to make target 'glide.yaml', needed by 'vendor'.  Stop.`

I was unclear of that until now so I will adding back

Makefile
sgraham785: I'm not sure which caching you are speaking about. It is really still doing the same thing I just made `$(EGOS)` run as a prerequisite of init target. the `glide.lock` caching still applies

sgraham785: also to elaborate more I chose to use 'init' over 'vendor' since it now also runs the ego complies and I feel more truly initiates the project now

Makefile
cainlevy: My understanding is that `vendor: glide.yaml` will run when a) the target is missing or b) the prerequisite is newer. Here's how it looks in my environment:

```bash
$ make vendor
make: `vendor' is up to date.
$ touch glide.yaml
$ make vendor
glide install
...
go install
$ make vendor
make: `vendor' is up to date.
```

sgraham785: what version on Make?

Mine:
```
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

Makefile
cainlevy: The difference, as I understand it, is that previously the prerequisites for this rule were actual files on disk and `make` was able to check mtime to see if the prerequisites were satisfied. That's academic though, as I've talked myself out of the caching.

> 'init' over 'vendor' since it now also runs the ego complies and I feel more truly initiates the project now

💯

Makefile
```diff
@@ -14,10 +14,11 @@ clean:
...
 	ego api/views
 
 # Fetch dependencies
-vendor: glide.yaml
+init: glide.yaml $(EGOS)
```
cainlevy: i think this needs to be specifically `vendor: glide.yaml` for make's caching to work. if you declare the `init` rule separately, we can have the best of both worlds.

sgraham785: hmm I see. so how about:

```
init: $(EGOS) vendor

# Fetch dependencies
vendor: glide.yaml
	glide install
	go install
```

Makefile
cainlevy: huh:

```
$ make -v
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

Makefile
cainlevy: looks good! i'll try it out when the branch is ready.

.travis.yml
```diff
@@ -9,7 +9,6 @@ before_install:
   - sudo add-apt-repository ppa:masterminds/glide -y
   - sudo apt-get update -q
   - sudo apt-get install glide -y
-  - go get github.com/benbjohnson/ego/cmd/ego
```
sgraham785: @cainlevy added this last minute

cainlevy:
thanks a bunch for smoothing over these issues

cainlevy:
fixes #40

sgraham785:
you are welcome. this now frees me to take on adding postgres ;)

PR #43, opened 2018-02-11
Update sqlite db.go

### LINKED ISSUE #42: Custom SQLite db path
Why is the db path fixed within current binary path?

https://github.com/keratin/authn-server/blob/d20f79cbdcd317e0cb2d80a3476b4901ed3af87e/data/sqlite3/db.go#L15

akhedrane:
Remove prefix (./) and suffix (.db) from formated connection uri.
this allow use of full path like (sqlite3://localhost/c:/full/path/to/app/authn.sqlite)

This PR will fix #42

### DIFF
--- a/data/sqlite3/db.go
+++ b/data/sqlite3/db.go
@@ -12,7 +12,7 @@ func NewDB(env string) (*sqlx.DB, error) {
 	// https://github.com/mattn/go-sqlite3/issues/274#issuecomment-232942571
 	// enable a busy timeout for concurrent load. keep it short. the busy timeout can be harmful
 	// under sustained load, but helpful during short bursts.
-	return sqlx.Connect("sqlite3", fmt.Sprintf("./%v.db?cache=shared&_busy_timeout=200", env))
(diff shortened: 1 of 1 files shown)


data/sqlite3/db.go
```diff
@@ -12,7 +12,7 @@ func NewDB(env string) (*sqlx.DB, error) {
 	// https://github.com/mattn/go-sqlite3/issues/274#issuecomment-232942571
 	// enable a busy timeout for concurrent load. keep it short. the busy timeout can be harmful
 	// under sustained load, but helpful during short bursts.
-	return sqlx.Connect("sqlite3", fmt.Sprintf("./%v.db?cache=shared&_busy_timeout=200", env))
+	return sqlx.Connect("sqlite3", fmt.Sprintf("%v?cache=shared&_busy_timeout=200", env))
```
cainlevy: Thanks for the patch!

My only hesitation here is backwards compatibility. What if we check to see whether the path contains a `.` and assume it's a newer style absolute path? Do you think that would work?

akhedrane: Ok, i think that would be perfect

cainlevy:
released in v1.2.1 🎉

PR #44, opened 2018-02-23
streamlined key rotation coordination

cainlevy:
Every AuthN server creates a single routine responsible for periodically rotating the current RSA key. Previously, this routine tried to coordinate with other servers through a Redis mutex. The mutex relied on a lock timeout that could be improperly configured and in the worst case yield competing keys and servers that could not verify each other's tokens.

In a single server environment, the coordination complexity and cost is pointless.

In a multi server environment, the coordination complexity is not worth the saved work. Each server can generate its own key and then compete to see whose key wins.

(Thanks to @shashankmehra for suggesting this enhancement)

### DIFF
--- a/data/blob_store.go
+++ b/data/blob_store.go
@@ -15,15 +15,7 @@ type BlobStore interface {
 	// Read fetches a blob from the store.
 	Read(name string) ([]byte, error)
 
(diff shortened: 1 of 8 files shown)


cainlevy:
On my laptop (i7 3.5ghz) I was averaging about 230ms to generate a key.

PR #50, opened 2018-05-08
Oauth

### LINKED ISSUE #11: OAuth support
Does this support the use-case where users signup / login via 3rd party OAuth (I.E Github / Facebook) ?

cainlevy:
Adds OAuth support to AuthN.

In this setup, AuthN hides all of the OAuth 2.0 implementation details from the main application. The main application redirects users through AuthN and then relies on the refresh endpoint to acquire an identity token.

An AuthN account may have a linked identity with each configured provider.

Only Google OAuth has been implemented, but other providers are easy to add.

Currently there is no way for the application to know if a user's session originated with OAuth. In the future, AuthN may provide the application with the user's current access token for business logic.

## Outcomes

There are three success outcomes of the OAuth integration:

1. Login -- if the provider identity has been linked before, AuthN will establish a session for the existing account. The application should already have a user profile.
2. Account Connection -- if the provider identity is unknown but AuthN has a valid session for an account that has not linked to the provider, then AuthN will connect the identity to the existing account.
3. Signup -- if the provider identity is unknown and AuthN does not have a current session, then AuthN will attempt to create a new account with the linked identity. *The application should detect a missing user profile and finish registration.*

Failure outcomes include:

* Attempting to redirect to a URL that does not match one of the configure application domains
* Tampering with the OAuth return URL
* Attempting a CSRF attack on the OAuth return action
* Taking more than an hour to complete the OAuth flow
* Attempting to authenticate with a locked account

## TODO

* [x] integration proof with demo app
* [x] Docs on new config variables
* [x] Docs on howto
* [x] More provider integrations (Facebook & GitHub are priorities)
* [ ] List the new feature (beta)

Fixes #11


k1ng440:
does it work now?

cainlevy:
yes! 🎉 

i currently have it working in a demo app, and have written some basic docs for how to integrate. i've been ill the past few days, though, and haven't settled on a release plan.

if you're in a position to build and try this branch that would be awesome. otherwise i'll plan on making a new release of authn with oauth in some kind of beta status.

cainlevy:
Hope this is helpful! https://github.com/keratin/authn-server/blob/b7bf46763b147b0b189db4ce0a431d5b1ed7f7ff/docs/guide-implementing_oauth.md

Issue #57, opened 2018-07-21
Dockerfile does not run migrate by default


ernsheong:
If one uses docker-compose from the get-go, this could be a gotcha. Should migrations generally be part of runtime or separate?

cainlevy:
I suppose that I defaulted to making them separate because they are not zero cost, even if they are fast. But you're right to observe that making them separate allows a dev to forget.

I'd like to hear a bit more about your experience. Do you think this is a gotcha primarily in development, or also in production? Would your instinct to bundle the migrations into startup be different in each environment?

It'd be nice to have an example docker-compose setup or guide for folks getting started. That might include a `command` override to ensure that migrations run in development. It wouldn't really help with production, though, where folks with Dockerized deployments are more likely to use ECS, Kubernetes, etc.

ernsheong:
Simply putting `command: ./authn migrate && ./authn server` doesn't work. Migration does, starting server doesn't. Maybe `command: ./authn migrate &; ./authn server;` might work

cainlevy:
@ernsheong try `command: sh -c "./authn migrate && ./authn server"`. i believe that's a docker limitation.

Mohammed90:
@cainlevy I've just used `command: sh -c "./authn migrate && echo 'starting server, migrations finished' && ./authn server"` in the `docker-compose.yml` with `DATABASE_URL=sqlite3://:memory:?mode=memory&cache=shared` and it works perfectly. I didn't think doing the two commands with `&&` would solve it 😄

PR #64, opened 2018-10-15
enable auto migrate

roylau134:
https://github.com/keratin/authn-server/issues/57#issuecomment-408477406


roylau134:
should i add auto migration to the docker build file so that everyone can easily use this app??

cainlevy:
@roylau134 I'm 👍 to document the approach as a "getting started" tip, but I'm not confident enough to make it the default behavior when folks use the Docker build. The main reason is that I'm not accustomed to applications or frameworks that automatically migrate a database. For example, Ruby on Rails migrations are a separate command from the server, even when the server is able to know that they are pending.

Issue #61, opened 2018-09-13
session storage with PostgreSQL


roychowdhuryrohit-dev:
I am running with the following error while connecting authn-server to my postgresql docker instance.

_panic: NewRefreshTokenStore: unsupported driver: postgres_

roychowdhuryrohit-dev:
I have found the source of the issue in NewRefreshTokenStore.go

`func NewRefreshTokenStore(db *sqlx.DB, redis *redis.Client, reporter ops.ErrorReporter, ttl time.Duration) (RefreshTokenStore, error) {
	if redis != nil {
		return &dataRedis.RefreshTokenStore{
			Client: redis,
			TTL:    ttl,
		}, nil
	}

	switch db.DriverName() {
	case "sqlite3":
		store := &sqlite3.RefreshTokenStore{
			DB:  db,
			TTL: ttl,
		}
		store.Clean(reporter)
		return store, nil
	default:
		return nil, fmt.Errorf("unsupported driver: %v", db.DriverName())
	}
}`

There's no support for Postgresql. Instead it only supports SQLite and Redis. Please help !

cainlevy:
Happy to reconsider if there's reason to prefer PostgreSQL over Redis.

PR #69, opened 2018-10-20
build a distributable for windows/amd64

### LINKED ISSUE #68: Build for Windows
Hey, @cainlevy 

Can you build a release of AuthN for Windows?

cainlevy:
Fixes #68 

@egorsmkv Please try https://github.com/keratin/authn-server/releases/download/v1.4.1/authn-windows64.bz2 and let me know how it goes!

* [x] Change dist naming so the upload to GH is less error-prone
* [ ] Extract port from AUTHN_URL
* [x] Active users page 404
* [ ] Support for `authn-server.env` file


egorsmkv:
I have ran it. Thanks for the build!

windows powershell 2018-10-20 18 35 29


Feedback:

1) The file https://github.com/keratin/authn-server/releases/download/v1.4.1/authn-windows64.bz2 is not a bz2 archive, it is an executable file.
2) By default, the port is not recognized, was just zero value. I changed this by set the PORT env variable.
3) The page "Active Users" does not work.
4) The process of configuring of env variables is not convenient method on Windows. Maybe add a method with a config file?

egorsmkv:
I wrote this Windows batch file for those of us who want to run AuthN on Windows:

Save this code as a file "run.bat" and run as an administrator. You need to have the "authn-windows64.exe" file where is this batch file and you need to specify the currentFolder variable which is a folder where you want to run AuthN.

```
set currentFolder=C:\Developement\authn

cd %currentFolder%

setx -m AUTHN_URL localhost:8080
setx -m APP_DOMAINS localhost
setx -m PORT 8080
setx -m SECRET_KEY_BASE my-authn-test-secret
setx -m DATABASE_URL "sqlite3://:memory:?mode=memory&cache=shared"

setx -m HTTP_AUTH_USERNAME test
setx -m HTTP_AUTH_PASSWORD test

%currentFolder%\authn-windows64.exe migrate
%currentFolder%\authn-windows64.exe

pause
```

cainlevy:
@egorsmkv the authn-server has https://github.com/joho/godotenv built in. does creating a `.env` file next to the binary make it easier?

egorsmkv:
It is not possible to create a file without a name on Windows.

cainlevy:
> By default, the port was not recognized from the AUTHN_URL env variable, it was just zero value. I changed this by set the PORT env variable.

🤔 

> The page "Active Users" does not work.

How does it fail?

> It is not possible to create a file without a name on Windows.

Ah, that makes sense. I sure picked the wrong time to wipe out my Windows partition!

Perhaps a quick fix here would be to support a Windows-friendly dotenv naming variation?

cainlevy:
It seems that my comment is late to the party. I'm 👍 on calling `godotenv.Load(...)` explicitly.

egorsmkv:
I have found a new problem. I tried to configure the DATABASE_URL variable for using MySQL. Migrations ran okay, but when I am starting the server there is this error:

```
NewRefreshTokenStore: unsupported driver: mysql
```

How can I see from the code: https://github.com/keratin/authn-server/blob/3e303ae8ca528438be5562bdfa141c158e003c9c/data/refresh_token_store.go#L39 there is must be redis or sqlite.

Did not implement this store yet?

egorsmkv:
> How does it fail?
Just 404 page.

> Perhaps a quick fix here would be to support a Windows-friendly dotenv naming variation?
I am testing the authn-server currently, do not make quick fixes. However, there is a way to run the authn-server like my batch file.

cainlevy:
> I am testing the authn-server currently, do not make quick fixes.

Understood.

> NewRefreshTokenStore: unsupported driver: mysql

Correct, this store currently requires Redis or Sqlite3. Similar discussion here: https://github.com/keratin/authn-server/issues/61

egorsmkv:
I see. I have set the REDIS_URL variable and it works well.

egorsmkv:
@cainlevy I added a user into the database via curl and the Active Users page now is working.

cainlevy:
I'm not sure why the server did not read `AUTHN_URL` for the default port assignment. I'd be happy to fix that as a follow-up bug report.

In the meantime, I'm going to merge this so that I can easily create Windows builds on future releases.

egorsmkv:
Maybe, this behavior was because I used global environment variables. I have rewrite own batch script https://github.com/keratin/authn-server/pull/69#issuecomment-431594816 and now the port from the AUTHN_URL variable detects okay.

PR #71, opened 2018-10-23
Password-less Login

etruta:
Possibility to login from a link (token).

This has 2 methods:
1. `GET /passwordless/token`: To request a passwordless token, that send the token to a webhook (responsible to send the email to account id);
2. `POST /passwordless/login`: To login with this token.

The webhook url is setted by env **APP_PASSWORDLESS_TOKEN_URL**.

This token can be used until expired, your time to expire is setted by env **PASSWORDLESS_TOKEN_TTL**.


cainlevy:
This is next on my list! I hope to get to it within the next few days. 🎉

api/passwordless/routing.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"github.com/keratin/authn-server/api"
+	"github.com/keratin/authn-server/lib/route"
+)
+
+func PublicRoutes(app *api.App) []*route.HandledRoute {
+	originSecurity := route.OriginSecurity(app.Config.ApplicationDomains)
+
+	routes := []*route.HandledRoute{
+		route.Post("/passwordless/login").
```
cainlevy: i'd like to merge these new endpoints into the `/sessions` api directory:

* `.Get("/sessions/token")`
* `.Post("/sessions/token")`

api/test/app.go
```diff
@@ -25,15 +25,16 @@ func App() *api.App {
...
-		EnableSignup:          true,
+		BcryptCost:              4,
+		SessionSigningKey:       []byte("TestKey"),
+		AuthNURL:                authnURL,
+		SessionCookieName:       "authn",
+		OAuthCookieName:         "authn-oauth-nonce",
+		ApplicationDomains:      []route.Domain{{Hostname: "test.com"}},
+		PasswordMinComplexity:   2,
+		AppPasswordResetURL:     &url.URL{Scheme: "https", Host: "app.example.com"},
+    AppPasswordlessTokenURL: &url.URL{Scheme: "https", Host: "app.example.com"},
```
cainlevy: indentation (did gofmt miss this one?)

config/config.go
```diff
@@ -248,6 +252,19 @@ var configurers = []configurer{
 		return err
 	},
 
+	// PASSWORDLESS_TOKEN_TTL determines how long a passwordless token (as JWT)
+	// will be valid from when it is generated. These tokens should not live much
+	// longer than it takes for an attentive user to act in a reasonably expedient
+	// manner. If a user loses control of a password reset token, they will lose
```
cainlevy:
```suggestion
	// manner. If a user loses control of a passwordless token, they will lose
```

api/passwordless/routing.go
etruta: do you think we should also change the function names to `getToken` and` postToken` or we keep the current names?

api/passwordless/post_passwordless_login.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/keratin/authn-server/api"
+	"github.com/keratin/authn-server/lib/route"
+	"github.com/keratin/authn-server/services"
+)
+
+func postPasswordlessLogin(app *api.App) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var err error
+		var accountID int
+		if r.FormValue("token") != "" {
```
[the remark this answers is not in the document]
etruta: wow really, detecting the error with `ErrInvalidOrExpired`, correct?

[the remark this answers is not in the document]
cainlevy: Yeah that sounds good

api/passwordless/routing.go
cainlevy: It'd be nice to keep parity between func names and routes.

What do you think about `getToken` and `postToken`? It's the best I could come up with on the spot, but I'm open to other ideas if you think it should be clearer.

etruta: maybe `getSessionToken` and `postSessionToken` to use the same pattern of `.Get("/session/refresh")` with `getSessionRefresh`.

etruta:
@cainlevy, What do you think about to update `accounts.last_login_at` in `api.NewSession`?

data/sqlite3/account_store.go
```diff
@@ -57,10 +57,11 @@ func (db *AccountStore) Create(u string, p []byte) (*models.Account, error) {
 		PasswordChangedAt: now,
 		CreatedAt:         now,
 		UpdatedAt:         now,
+		LastLoginAt:       now,
```
cainlevy: can you say more about why this was needed for sqlite3 but not mysql or postgresql?

data/testers/account_store_testers.go
```diff
@@ -241,3 +241,18 @@ func testFindByOauthAccount(t *testing.T, store data.AccountStore) {
 	// Assert that db connections are released to pool
 	assert.Equal(t, 1, getOpenConnectionCount(store))
 }
+
+func testSetLastLogin(t *testing.T, store data.AccountStore) {
```
cainlevy: you'll want to add this func to the list of testers at the top of the file

services/account_last_login.go
```diff
@@ -0,0 +1,18 @@
+package services
+
+import (
+	"github.com/keratin/authn-server/data"
+	"github.com/pkg/errors"
+)
+
+func LastLoginUpdater(store data.AccountStore, accountID int) error {
+	account, err := store.Find(accountID)
```
cainlevy: i'm wondering if we can optimize this down to a single query, from `SELECT` + `UPDATE` to simply `UPDATE`.

* mysql should return the number of affected rows
* postgresql can do something similar with `RETURNING`
* sqlite probably requires a second query

so maybe each store should be responsible for returning true/false if the operation was successful, in addition to any possible err?

```go
type AccountStore interface {
	SetLastLogin(id int) (bool, error)
}
```

services/account_last_login_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/keratin/authn-server/services"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/keratin/authn-server/data/mock"
+)
+
+func TestLastLoginUpdater(t *testing.T) {
+	t.Run("last login", func(t *testing.T) {
```
cainlevy: would you add a test for the ErrNotFound behavior? i don't mind if it's part of this `t.Run` or in another.

services/passwordless_token_verifier.go
```diff
@@ -32,6 +32,8 @@ func PasswordlessTokenVerifier(store data.AccountStore, r ops.ErrorReporter, cfg
 		return 0, FieldErrors{{"account", ErrLocked}}
 	} else if account.Archived() {
 		return 0, FieldErrors{{"account", ErrLocked}}
+	} else if account.LastLoginAt.Sub(claims.IssuedAt.Time()) > 0 {
```
cainlevy: ahhh, nice, we don't need a `Lock` claim after all. good find!

would `Before` work here?

```suggestion
	} else if account.LastLoginAt.Before(claims.IssuedAt.Time()) {
```

data/mysql/migrations.go
```diff
@@ -28,6 +28,7 @@ func createAccounts(db *sqlx.DB) error {
             locked TINYINT(1) NOT NULL DEFAULT '0',
             require_new_password TINYINT(1) NOT NULL DEFAULT '0',
             password_changed_at DATETIME DEFAULT NULL,
+            last_login_at DATETIME DEFAULT NULL,
```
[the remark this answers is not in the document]
etruta: Ow really, to work in old installations. In this case, I will need to check if the field exist. In postgres this works fine with `ADD COLUMN IF NOT EXISTS`, but in mysql i need to check `INFORMATION_SCHEMA.COLUMNS` and know the database name in migrations. Do you know/advise another way in mysql?

data/sqlite3/account_store.go
etruta: Ops, my bad, the correct is mysql and postgresql way. I need only change to `LastLoginAt` to work with nil value, like `DeletedAt`.

data/mysql/migrations.go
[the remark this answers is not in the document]
cainlevy: In MySQL I'd probably just catch the error if the column already exists!

cainlevy:
@etruta I have a bit of extra time coming up. Let me know if I can be of any help?

etruta:
Hi @cainlevy, it's ok. I returned today with more time to this project too, until tomorrow I hope to solve this questions.

services/passwordless_token_verifier.go
etruta: Would not `After`? Returns the `ErrInvalidOrExpired` error if exists a login (`LastLoginAt`) after request a passwordless token (`claims.IssuedAt`) ?

PR #83, opened 2018-11-16
gRPC Support

### LINKED ISSUE #78: convert to gRPC with REST proxy
AuthN could support both gRPC and REST by converting the core service to gRPC and integrating https://github.com/grpc-ecosystem/grpc-gateway. This could make it easier to write backend clients in target languages while maintaining flexibility for frontend clients that don't want to bundle grpc-web (50kb).

Mohammed90:
This PR introduces gRPC support with Protobuf. 

Closes #78 

Continuous feedback is preferred instead of waiting for me to finish.

Tasks:

- [x] refactor existing code so it can be reused more readily between HTTP and gRPC servers
- [ ] remove the current server into authn-server/http
- [x] work on a parallel authn-server/gRPC package that can be preferred with a command flag
- [ ] if and when appropriate, bring in grpc-gateway to replace the authn-server/http package
- [x] protect the gRPC API from abuse (existing REST API validates Origin header, gRPC does .... [insert something here])
- [ ] tests, tests, tests
- [N/A] Fix mTLS with with cmux & add tests for it (I feel this is going to be a giant test)
- [x] Address comments [here](https://github.com/keratin/authn-server/pull/83#discussion_r272802380). It's pretty tricky to eliminate dependency on vendor directory. 

Critical bug in current branch:
- [x] mounting on a subpath not working. My hunch is wrapping the handler with `http.StripPrefix` should fix it, but I'm yet to tackle it.
- [x] Work around [this "bug"](https://github.com/grpc-ecosystem/grpc-gateway/issues/920).

Resources used and/or might need as reference to complete PR:

- [So you want to use GoGo Protobuf](https://jbrandhorst.com/post/gogoproto/)
- [Authentication in gRPC](https://mycodesmells.com/post/authentication-in-grpc)
- [HTTP server from gRPC with Gateway](https://mycodesmells.com/post/http-server-from-grpc-with-gateway)
- [gRPC-Gateway Error Handler](https://mycodesmells.com/post/grpc-gateway-error-handler)
- [Advanced gRPC Error Usage](https://jbrandhorst.com/post/grpc-errors/)
- [Workaround to marshal application/x-www-form-urlencoded w/ grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/issues/7#issuecomment-358569373)
- [https://github.com/grpc-ecosystem/go-grpc-prometheus](https://github.com/grpc-ecosystem/go-grpc-prometheus)


grpc/authn.proto
```diff
@@ -0,0 +1,246 @@
...
+  string e = 5;
+  string n = 6;
+}
+message JWKSResponse { repeated Key keys = 1; }
+
+service AuthN {
+
+  rpc Signup(SignupRequest) returns (SignupResponseEnvelope) {
+    option (google.api.http) = {
+      post : "/accounts",
```
cainlevy: before this is finished, we should check how the gateway handles being mounted on a subpath of the domain (e.g. `mydomain.com/authn`).

we'll also need to consider what this means for public and private routing tables.

Mohammed90:
For the first point, I have 2 ideas from the top of my head:

1. Instead of 1 service in the proto file for the entire application, e.g. PrivateAuthN and PublicAuthN. The gRPC server is then wrapped with an interceptor that checks strings#HasPrefix "/grpc.PrivateAuthN". If true,  a form of authentication is applied (e.g. Basic Auth or TLS Mutual Auth); otherwise the request is passed immediately.

1. Maintain a map of the private routes. An interceptor checks the requested route. If the route is in the map, then the request must be authenticated.

The first approach has the advantage of being low-maintenance and not prone to missing out some routes. Again, those are from the top of my head solutions and may not be the best. More research could reveal cleaner approaches.

Looking more into the generated and example code, it seems that the gateway listens to HTTP on a different port, calls the corresponding gRPC method, and hands back the response to the HTTP request. Given that, it's possible to not use grpc-gateway and just listen on the different ports. But this will cause duplication of logic between the 2 listeners. Using the gateway will avoid the duplicate logic but will require serious refactor of the API package.

cainlevy:
I'd like to think about ways of approaching gRPC gradually and incrementally. The high level strategy here is to run gRPC with a HTTP gateway, but I wonder if there's a way to instead make it an optional (unofficial) build until it's mature?

1. refactor existing code so it can be reused more readily between HTTP and gRPC servers
2. move the current server into `authn-server/http`
3. work on a parallel `authn-server/gRPC` package that can be preferred with a command flag
4. if and when appropriate, bring in grpc-gateway to replace the `authn-server/http` package

What do you think? I'd be happy to work on steps 1 and 2, myself, if you had any recommendations for how to accomplish better reuse.

Mohammed90:
Dang, that `GET /metrics` in particular is an annoying obstacle. 

I can't imagine how to use grpc-gateway to keep the HTTP endpoints without having gRPC endpoint as well (i.e. gRPC being optional) because the HTTP server (from the grpc-gateway) internally calls the gRPC server [(see here)](https://github.com/keratin/authn-server/blob/607a9943e61c49470df0277e502cfbf6e52d367b/grpc/authn.pb.gw.go#L411). Perhaps during the transition period, a flag will select whether to only enable the HTTP server (current implementation) or the gRPC server which includes the HTTP API.

The proposed approach looks good. I can't think of any improvements on it. As you've suggested, you'll do better on the refactor (steps 1 and 2) because you're more familiar with the codebase than I am. I'll take up the gRPC part. I'll start by splitting the AuthN service into 2 services, public service and administration service, so they can listen on different ports.

cainlevy:
I'd like to return to an earlier question: why not use the authn-js client to communicate directly with the AuthN HTTP handlers?

Using gRPC from the browser requires grpc-web, which is a different protocol that requires a backend proxy to communicate with gRPC services. This adds indirection and an extra point of failure for password security (e.g. accidental logging).

I've read https://improbable.io/games/blog/grpc-web-moving-past-restjson-towards-type-safe-web-apis and the benefits of grpc-web in their environment sound amazing. I'm a bit envious of the type-safe and automated integration between their backend and frontend.

Those benefits don't apply to AuthN, though, because AuthN is committed to maintaining a client for compatibility with non-gRPC frontend environments. The authn-js client exists and must continue to exist. It was hand-crafted, but is finished and already provides type safety with bundled TypeScript definitions.

So what benefits do you expect from AuthN supporting gRPC-web, compared to using the official authn-js client?

(Regardless of your answer to that question, I'm still interested in gRPC for "private" endpoints.)

Mohammed90:
gRPC-Web support is just a side-effect of having a gRPC API 😄AuthN currently is tedious to use for mobile clients (apps) because it expects to be called from a browser by expecting the Origin header. Mobile clients have to fake being a browser by setting the Origin header. If we keep going on the current path of this PR, current authn-js users can keep using the library thus calling the HTTP interface, gRPC-web users will be able to use the gRPC API, and mobile clients can use AuthN without having to fake being a browser.

Unless I misunderstood your question...

cainlevy:
> mobile clients can use AuthN without having to fake being a browser

Oooo, that's what I was missing! I didn't realize this was solving for non-browser clients. Yes, it makes sense now that `Origin` and `Cookie` headers would be problems. I've been expecting this moment but didn't recognize it when it arrived. 😅

I'll reach out on gitter to catch up more quickly on the topic.

grpc/authn-private.proto
```diff
@@ -0,0 +1,106 @@
...
+
+service AdminAuthN {
+
+  rpc GetAccount(GetAccountRequest) returns (GetAccountResponseEnvelope) {
+    option (google.api.http) = {
+      get : "/accounts/{id}",
+    };
+  }
+
+  rpc UpdateAccount(UpdateAccountRequest) returns (google.protobuf.Empty) {
```
aybabtme: Sorry this is totally unsolicited feedback, but may I suggest that you use dedicated response types for each RPC, including those that return nothing? The types of your RPC request/response become part of the API, and later changing them (say you decide to add a value in the response) would mean breaking the API in a non-backward compatible manner.

So in general, it's preferable to have all RPC look like:

```protobuf
service Blah {
    rpc DoThing(DoThingRequest) returns (DoThingResponse);
}

message DoThingRequest {
    // stuff goes here
}

message DoThingsResponse { } // empty message
```
Also note that each RPC should have it's own `Request`  and  `Response` types, and not share them between RPCs (for the same reasons).

grpc/authn-private.proto
```diff
@@ -0,0 +1,106 @@
+syntax = "proto3";
+
+package grpc;
```
aybabtme: More unsolicited feebback, sorry! A package name like `keratin/authn` will lead to better looking code in users calling your server. Also, it will prevent collisions in the types, because for anyhow using your gRPC client, there's a global namespace of Protobuf types that are registered by `$package_name` and `$message_name`. So if your package name is just `grpc`, that's more likely to collide with other protobuf types that your users have in their apps.

aybabtme:
With regard to using 4 listeners, you can use [`cmux`](https://github.com/soheilhy/cmux) to multiplex requests on a single listener to either HTTP or gRPC backends. See an example service that does it here: https://gist.github.com/soheilhy/bb272c000f1987f17063#file-server-go-L265-L284

CockroachDB [used that for a while](https://www.cockroachlabs.com/blog/a-tale-of-two-ports/) (not sure if they still do).

Mohammed90:
Thank you for the review, @aybabtme! Reviews are more welcome now than later exactly to not cause a hassle of breakage. Can I ask you to stick around this PR or do another round of review later? I'll address your current comments in my upcoming commits.

grpc/internal/gateway/gateway.go
```diff
@@ -0,0 +1,89 @@
...
+type GatewayResponseMiddleware func(ctx context.Context, response http.ResponseWriter, m proto.Message) error
+
+// StatusCodeMutator changes the HTTP Repsonse status code to the desired mapping. The default mapping
+// by gRPC-Gateway doesn't have some of the desired responses (e.g. 201), which is why this function is needed.
+func StatusCodeMutator(ctx context.Context, response http.ResponseWriter, m proto.Message) error {
+	log.Printf("%s: message type: %T", time.Now(), m)
+	switch m.(type) {
+	case *authnpb.SignupResponseEnvelope, *authnpb.LoginResponseEnvelope, *authnpb.RefreshSessionResponseEnvelope, *authnpb.SubmitPasswordlessLoginResponseEnvelope, *authnpb.ChangePasswordResponseEnvelope:
+		response.WriteHeader(http.StatusCreated)
+	}
```
Mohammed90: Perhaps I should fallback to `runtime.HTTPStatusFromCode(code codes.Code)` here?

grpc/internal/gateway/gateway.go
```diff
@@ -0,0 +1,89 @@
...
+			r.Body = ioutil.NopCloser(bytes.NewReader(jsonBody))
+			r.ContentLength = int64(len(jsonBody))
+			r.Header.Set("Content-Type", "application/json")
+		}
+		mux.ServeHTTP(w, r)
+	})
+}
+
+// HeaderPrinter logs the headers of the request
+func HeaderPrinter(mux http.Handler) http.Handler {
```
Mohammed90: This was used for debugging. I should remove it.

grpc/private/routing.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/keratin/authn-server/api"
+	"github.com/keratin/authn-server/grpc/public"
+	"github.com/keratin/authn-server/lib/route"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+/*
+func router(app *api.App) http.Handler {
+	r := mux.NewRouter()
```
Mohammed90: I used this comment block while writing this file to constantly see what routes I should check and add. It should be removed or add a comment referencing where these came from.

grpc/public/public_server.go
```diff
@@ -0,0 +1,312 @@
...
+type accountIDKey int
+
+type publicServer struct {
+	app *api.App
+}
+
+func RunPublicGRPC(ctx context.Context, app *api.App, l net.Listener) error {
+	srv := grpc.NewServer(
+		grpc_middleware.WithUnaryServerChain(
+			logInterceptor,
```
Mohammed90: Should be removed in favor of `go-grpc-middleware/logging/logrus`

silasdavis:
I am also need gRPC support on the front and backend. I'd also like to start testing with it as soon as possible. @Mohammed90 is there anything I could help with?

Mohammed90:
@silasdavis, unfortunately, the remaining work cannot be divided any further. It has to be completed by me. I'm currently 80% through writing the tests, and that will be followed by wiring the gRPC portion to the main func. You can help me with code review, if that's okay with you. 
Later on, it'd also be great if you can test it before merging the PR, but this depends on whether you already have an app/infrastructure to test with 😃

Mohammed90:
Fellas (@cainlevy , @aybabtme , @silasdavis ), the implementation is now complete. Could you please review it? I know there are bits of cleanup to be handled (review the gRPC error codes returned and ensure they're consistent, e.g. FailedPrecondition and NotFound); meanwhile, it'd be great to have a few pairs of eyes to look at the implementation.

Enabling gRPC is hidden behind `EXPERIMENTAL` env var, so you need to set this to any of the accepted boolean values in addition to setting gRPC specific env vars.

P.S.: The coverage decrease bothers me so much, but most of it is due to generated code for gRPC-gateway.

Makefile
```diff
@@ -101,3 +102,26 @@ release: test dist
 	git push --tags
 	open https://github.com/$(NAME)/releases/tag/v$(VERSION)
 	open dist
+
+.PHONY: generate-grpc
+generate-grpc:
```
cainlevy: This make script depends on a few unstated things:

* the `protoc` binary
* the vendor/ directory (from `go mod vendor`?)
* the `github.com/grpc-ecosystem/grpc-gateway` package (so that `GATEWAY_SRC` resolves)
* i'm currently blocked on
  > protoc-gen-gogoslick: program not found or is not executable`

Which of those things can be resolved automatically in the Makefile using a dependency, and which need to be documented in the README?

grpc/server/server.go
```diff
@@ -0,0 +1,135 @@
...
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/grpc"
+)
+
+func init() {
+	runtime.HTTPError = errors.CustomHTTPError
+}
+
+// RunPrivateService starts a gRPC server for the private API and accompanying gRPC-Gateway server
+func RunPrivateService(ctx context.Context, app *app.App, grpcListener net.Listener, httpListener net.Listener) error {
```
cainlevy: this func does not appear to be used outside the package. minimizing exported functions helps with code comprehension. would you review other packages as well?

```suggestion
func runPrivateService(ctx context.Context, app *app.App, grpcListener net.Listener, httpListener net.Listener) error {
```

grpc/server/server.go
```diff
@@ -0,0 +1,135 @@
...
+
+	g.Go(func() error {
+		return private.RunPrivateGateway(ctx, app, privateRouter, privClientConn, httpListener)
+	})
+
+	return g.Wait()
+}
+
+// RunPublicService starts a gRPC server for the public API and accompanying gRPC-Gateway server
+func RunPublicService(ctx context.Context, app *app.App, grpcListener net.Listener, httpListener net.Listener) error {
```
cainlevy:
```suggestion
func runPublicService(ctx context.Context, app *app.App, grpcListener net.Listener, httpListener net.Listener) error {
```

app/config.go
```diff
@@ -507,6 +513,62 @@ var configurers = []configurer{
...
+			val, err := lookupBool("ENABLE_GRPC", false)
+			if err == nil {
+				c.EnableGRPC = val
+			}
+			return err
+		}
+		return nil
+	},
+
+	// TLS_CERT and TLS_KEY pair for gRPC server
```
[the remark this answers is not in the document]
Mohammed90: This is for gRPC [mutual-TLS authentication](https://grpc.io/docs/guides/auth.html#with-server-authentication-ssltls).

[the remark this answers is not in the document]
cainlevy: What's gained from authenticating a client's cert? Is this intended to replace the `Origin` check on REST requests?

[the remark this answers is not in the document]
Mohammed90: Yes. It's also optional in our case, which I should have documented, so local development can be done with these values left empty. If both of these env vars are set, then the gRPC server will authenticate the clients with mTLS, otherwise its served `.WithInsecure()`. See:

https://github.com/keratin/authn-server/blob/f71d462bd316d2f599c467030df43b0436161cba/grpc/server/server.go#L41-L50

Mohammed90:
Something is up with gcc in Travis...

app/config.go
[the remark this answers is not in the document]
cainlevy: Thanks for pointing that out, it does appear to make development easier.

Now my primary concerns are managing code/configuration complexity and understanding choices that I'll be responsible for maintaining. Please bear with me, I'm coming from a different operating environment and may not have the same context as you do.

I'm still uncertain what's gained from authenticating a gRPC client. The `Origin` check exists only to prevent XSRF attacks. Is that a concern with gRPC?

Or is this more of a replacement for HTTPS Basic Auth to secure communication between backends?

If AuthN didn't support mTLS, what would that mean for your deployment topology? Do you (would you) have a load balancer/proxy/sidecar that terminates TLS?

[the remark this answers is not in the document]
silasdavis: A replacement for basic auth and TLS termination - if you happen to be deploying client certificates this means you can dispatch with TLS terminating sidecars altogether.

grpc/public/password_reset.go
```diff
@@ -0,0 +1,32 @@
...
+
+type passwordResetServer struct {
+	app *app.App
+}
+
+func (s passwordResetServer) RequestPasswordReset(ctx context.Context, req *authnpb.PasswordResetRequest) (*authnpb.PasswordResetResponse, error) {
+
+	account, err := s.app.AccountStore.FindByUsername(req.GetUsername())
+	if err != nil {
+		panic(err)
```
silasdavis: Should this really be fatal? What if my `AccountStore` is temporarily unavaillable?

grpc/public/passwordless.go
```diff
@@ -0,0 +1,73 @@
...
+		s.app.Reporter,
+		s.app.Config,
+		req.GetToken(),
+	)
+
+	if err != nil {
+		if fe, ok := err.(services.FieldErrors); ok {
+			return nil, errors.ToStatusErrorWithDetails(fe, codes.FailedPrecondition).Err()
+		}
+		panic(err)
```
silasdavis: As another example, this looks like a sanity panic, i.e. if we get anything other then `FieldErrors` we panic - really?

Mohammed90:
@silasdavis those are 1-to-1 translated from current handlers code residing in the directory server/handlers.

silasdavis:
Okay sorry I should have read this first:

> ### Error Handling

> Errors are passed up from services to handlers. If an error is not recoverable, the handler should
panic (fatal) or report (warning). Panics are reported and generate a 500 error.

So it does seem like this is consistent. However I would question this use of panic. I don't think these are truly unrecoverable errors for the runtime, and in my experience using panic/recover in Go for control flow ends in tears...

Anyway not an issue for this PR.

app/config.go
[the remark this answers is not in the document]
cainlevy: Suppose some day that AuthN backend clients are updated to use gRPC and the admin REST endpoints are deprecated. Would that make client certs required for authenticating between backends?

Separately, suppose that AuthN is responsible for terminating TLS with a frontend client. Does that mean there's no load balancer, and if so, how could AuthN scale beyond a single instance?

Mohammed90:
Grrr.. cmux breaks mTLS authentication somehow. It works when I don't use it and use different ports instead. I'll work on finding a fix.
Current TODO:
[ ] Fix mTLS with with cmux & add tests for it (I feel this is going to be a giant test)
[ ] Address comments [here](https://github.com/keratin/authn-server/pull/83#discussion_r272802380). It's pretty tricky to eliminate dependency on vendor directory. 

I'll add them to OP as well.

app/config.go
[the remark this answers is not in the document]
silasdavis: @Mohammed90 yeah that probably makes sense, in principle you could dispense with the per request authz via Basic Auth if you chose to make that implicit with the client cert, but you could have a situation where a service has access to the client cert but isn't mean to make admin requests to the admin interface. It's also more consistent.

> if the gRPC service is served insecured

I don't think anyone is suggesting it be unsecured. For public gRPC would need to TLS terminate the connection at a proxy/load balancer (just like we to with the public HTTP interface) - gRPC uses HTTP/2 so we can do this with a proxy that supports it. I think the main point of enabling TLS would be for service-to-service backend communication. Now there we could also use a service mesh thingy to terminate the TLS, but it seems like less hassle and more supported. The setup I would have would be mTLS on private backend connection and a load balancer terminating the gRPC connection for public connection (see for example https://blog.getambassador.io/securing-grpc-connections-with-tls-ssl-and-ambassador-bdd3ea0396c6).

Alternatively you could take the same approach as with HTTP as define it as out-of-scope for authn-server. Something like Istio can give you the mTLS on the backend and Envoy on the front end, so it would work. I'd be tempted to support direct TLS on both transports but I see why you haven't on HTTP. I'd say the difference for gRPC relates to the different practice around gRPC and maybe the availability of suitable h2c proxies, which you might not bother with in the backend.

Makefile
silasdavis: Regarding the protobuf dependencies in vendor, can I suggest we just copy the ones we need into the repo (e.g. https://github.com/hyperledger/burrow/tree/develop/protobuf). I realise this feels a bit icky but:

- It makes it much easier for other languages to consume our protobuf - particular javascript which until recently couldn't even handle multiple import roots. It's a pain for a python or C++ app to have to care about go tooling to get vendor
- Just because a Go dependency depends on a particular protobuf version it doesn't automatically mean we do (though it kind of does for grpc-gateway I suppose)
- These google protobuf types are very stable - we realistically will not have to worry about updating them any time soon

We can include a make target to copy them in to our GRPC as a middle ground...

app/config.go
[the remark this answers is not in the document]
Mohammed90: Let's clear things up a bit first, this is for [Mutual Authentication](https://en.m.wikipedia.org/wiki/Mutual_authentication), not for serving the service on HTTPS. Different layers.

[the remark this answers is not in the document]
silasdavis: >Separately, suppose that AuthN is responsible for terminating TLS with a frontend client. Does that mean there's no load balancer, and if so, how could AuthN scale beyond a single instance

There is nothing to stop you from terminating TLS at multiple instances of authn all with the correct cert. Though it doesn't work for retries because the load balancer can't decrypt the connection and get the status code (e.g. 500).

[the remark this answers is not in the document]
Mohammed90: Okay, I'll eat my words. Even the load-balancer/reverse-proxy needs to authenticate with the certs if the gRPC server is configured as such, which makes sense in retrospect 😓 I guess I'm open to whatever decision you come up with for this.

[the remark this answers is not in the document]
cainlevy: I'd like to consider this functionality in some future issue or pull request. When features can be separated, that's my preference. It sounds like the gRPC API can still achieve the same security as the REST API without it.

One advantage to approaching this later is that we might consider it more holistically. If AuthN begins to assume responsibility for connection security, should it also affect the REST API? Public, private, or both? Those questions will be easier to discuss when the answers don't feel like scope creep in an already meaningful changeset.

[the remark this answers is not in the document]
Mohammed90: Ok, I've removed the TLS configuration.

grpc/internal/meta/session.go
```diff
@@ -0,0 +1,94 @@
...
+			lookupOnce.Do(func() {
+				var err error
+				session := parse()
+				if session == nil {
+					return
+				}
+
+				accountID, err = app.RefreshTokenStore.Find(models.RefreshToken(session.Subject))
+				if err != nil {
+					app.Reporter.ReportError(pkgerrors.Wrap(err, "Find"))
```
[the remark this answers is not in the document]
Mohammed90: Using `LogReporter` implementation of `app.Reporter.ReportRequestError` as example, perhaps I can add `ReportGRPCError(err error, info *grpc.UnaryServerInfo, req interface{})` to the `ErrorReporter` interface, and the LogReporter implementation of it can be something like this:

```go
func (r *LogReporter) ReportGRPCError(err error, info *grpc.UnaryServerInfo, req interface{}) {
	log.Printf("[%v][%T %v] [%T] %v\n", time.Now(), info.Server, info.FullMethod, req, err)
}
```

This will log which server implementation had the error and which method is the target, in addition to the error itself.

How does this sound? Printing the `req` can be harmful as it can contain passwords, so only the type is reasonable. The metadata stored in the context can also be sensitive because it can have the Basic Auth data.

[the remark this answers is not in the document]
cainlevy: Yes, that looks good. I defer to your judgement on what details are useful for debugging a GRPC error. Thanks for making sure to not print passwords or sessions.

The SentryReporter and AirbrakeReporter implementations can probably satisfy the new interface by passing the error along to `ReportError`, without any extra context.

Makefile
Mohammed90: Okay, I was able to remove dependency on `vendor/`. Regarding the dependency on `protoc` and the other binaries, a decision can be made between:

- document the dependencies in the README
- use the [znly/protoc](https://hub.docker.com/r/znly/protoc) docker image

The first one has the advantage of owning our dependency, kinda, and being able to always be on the latest version, but they're a few additional stuff to install. The Docker image route requires no additional installs, but it's a dependency on an external Docker image and we don't own the update cycle.

Opinions?

grpc/internal/meta/session.go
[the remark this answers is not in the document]
Mohammed90: I've added the method. I added context information for Airbrake and Sentry because I think it'll be confusing without them given there are multiple services in the application (although I could be completely wrong because I'm not familiar with Sentry and/or Airbrake).

I'll play with a Sentry docker container to see how the current implementation of reporting plays out with the overall implementation.

Mohammed90:
To keep you all on the same page, I'm currently writing an integration test that cross-verifies successful API calls across interfaces. It has already uncovered 2 bugs, one of them is the JSON encoding that's fixed in 21b06d9.

Makefile
cainlevy: I'm inclined to use a Docker image with all of the tooling necessary to generate these files. That seems consistent with how this Makefile currently relies on Docker when it needs a specific build environment.

We can always swap out one image for another (even our own) if need arises.

Mohammed90: Coincidentally, I made the same decision for the same reason last weekend :) Working on the necessary changes now

Mohammed90: Done!

Mohammed90:
In terms of functionality, this PR is complete. Until the review process is complete, I'll be working on:
- Explore Sentry to be familiar with it and find what gRPC specific metadata I can include
- Collect measurements to figure out sane {Read,Write,Idle}Timeouts for the server
- add more tests
- and definitely attend to review comments

grpc/private/gateway.go
```diff
@@ -0,0 +1,83 @@
...
+		runtime.WithForwardResponseOption(gateway.CookieSetter(app.Config)), // Cookies always have to go first
+		runtime.WithForwardResponseOption(gateway.StatusCodeMutator),
+		// Workaround this limitation: https://github.com/grpc-ecosystem/grpc-gateway/issues/920.
+		// Go's JSON encoder doesn't convert (u)int64 to strings silently.
+		runtime.WithMarshalerOption(runtime.MIMEWildcard, gateway.JSONMarshaler()),
+		runtime.WithMetadata(gateway.CookieAnnotator(app)),
+	)
+
+	r.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {
+		if req.URL.Path == "/" {
```
cainlevy: this looks like the only difference between here and `handlers.GetRoot`. why is it necessary? is it relevant to both locations?

grpc/private/gateway.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/keratin/authn-server/grpc/public"
+	"github.com/keratin/authn-server/server/views"
+	"golang.org/x/net/context"
+	grpc "google.golang.org/grpc"
+)
+
+func RunPrivateGateway(ctx context.Context, app *app.App, r *mux.Router, conn *grpc.ClientConn, l net.Listener) error {
+
+	gmux := runtime.NewServeMux(
+		runtime.WithForwardResponseOption(gateway.CookieSetter(app.Config)), // Cookies always have to go first
```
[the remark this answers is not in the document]
Mohammed90: Yes

grpc/private/gateway.go
Mohammed90: My bad. `handlers.GetRoot` can definitely be used.

Mohammed90:
@cainlevy , since you're going to play with a local server,  can you do me a favor? Could you collect some measurements and come up with reasonable timeouts for the `http.Server{}` for the servers in `publice/gateway.go` and `private/gateway.go`? It's discouraged to have them unset.

I'm planning on adding benchmark tests using [https://github.com/emicklei/hazana](https://github.com/emicklei/hazana) to figure out those numbers, but I'm not sure how long it'll be before I get around to that.

grpc/authn.proto
```diff
@@ -0,0 +1,168 @@
...
+}
+
+message SignupResponse {
+  string id_token = 1 [ (gogoproto.jsontag) = "id_token" ];
+}
+
+message SignupResponseEnvelope { SignupResponse result = 1; }
+
+message IsUsernameAvailableRequest { string username = 1; }
+message IsUsernameAvailableResponseEnvelope {
```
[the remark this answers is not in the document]
Mohammed90: The word `envelope` is used to keep the nomenclature consistent with the REST API as described [in the docs here](https://keratin.github.io/authn-server/#/api?id=json-envelope).

grpc/authn-private.proto
```diff
@@ -0,0 +1,173 @@
...
+message GetAccountRequest { int64 id = 1; }
+
+message GetAccountResponse {
+  int64 id = 1;
+  string username = 2;
+  bool locked = 3 [(gogoproto.jsontag) = "locked" ];
+  bool deleted = 4 [(gogoproto.jsontag) = "deleted" ];
+}
+
+message GetAccountResponseEnvelope { GetAccountResponse result = 1; }
```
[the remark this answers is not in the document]
Mohammed90: Such structure is used when the response contains a JSON nested inside the response JSON and keyed by with `result`.

```json
{
    "result": {
        // stuff
    }
}```

File: Makefile [hunk 19]
```diff
@@ -101,3 +125,26 @@ release: test dist
...
+	Mgoogle/protobuf/wrappers.proto=github.com/gogo/protobuf/types,plugins=grpc:./grpc \
+	--grpc-gateway_out=\
+	Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types,\
+	Mgoogle/protobuf/duration.proto=github.com/gogo/protobuf/types,\
+	Mgoogle/protobuf/empty.proto=github.com/gogo/protobuf/types,\
+	Mgoogle/api/annotations.proto=github.com/gogo/googleapis/google/api,\
+	Mgoogle/protobuf/field_mask.proto=github.com/gogo/protobuf/types:./grpc \
+	grpc/authn.proto grpc/authn-private.proto
+	# Workaround for https://github.com/grpc-ecosystem/grpc-gateway/issues/229
+	sed -i.bak "s/empty.Empty/types.Empty/g" grpc/authn.pb.gw.go && rm grpc/authn.pb.gw.go.bak
```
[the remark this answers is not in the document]
Mohammed90: We used to have `Empty` during the young life of this PR, then they were removed in favor of our own empty messages. I left this here in case `Empty` shows up in the future for whatever reason and we need this hack.

grpc/internal/gateway/router.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/keratin/authn-server/server/sessions"
+)
+
+// WrapRouter returns a handler with the following middlewares:
+// - Combined Logging (in Apache Combined Log Format)
+// - Sessions managing middleware
+// - CORS
+// - Proxy headers (if applicable)
+// -  Panic handler
+func WrapRouter(r http.Handler, app *app.App) http.Handler {
```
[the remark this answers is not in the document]
Mohammed90: Possible, and will eliminate this duplication, but the type of the first parameter needs to change from `*mux.Router` to `http.Handler`. It doesn't seem like a breaking change.

grpc/internal/errors/errors.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/grpc-ecosystem/grpc-gateway/runtime"
+	"github.com/keratin/authn-server/app/services"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+const wwwAuthneticate = `WWW-Authenticate`
+
+type FieldError struct {
```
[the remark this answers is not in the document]
Mohammed90: My memory is fuzzy on the details that lead me down this path. I presume it's because `services.fieldError` is unexported, and we need to convert gRPC's `BadRequest_FieldViolation` back to `services.FieldErrors`. I'll export `services.fieldError`, change the code, and see how it turns up.

grpc/private/gateway.go
Mohammed90: There's something awkward with the root path due to an idiosyncrasy of Gorilla's `PathPrefix`. Whenever `PathPrefix` is used, StrictSlash is automatically set to `true`. So if the `AUTHN_URL` is `/authn`, then it can only be reached on `/authn/` and never `/authn`. FWIW, this is the current behavior on `master` anyways.

grpc/internal/errors/errors.go
```diff
@@ -0,0 +1,113 @@
...
+// CustomHTTPError is a custom error handler to write the error JSON obeying the pre-defined error JSON structure
+func CustomHTTPError(ctx context.Context, mux *runtime.ServeMux, marshaler runtime.Marshaler, w http.ResponseWriter, req *http.Request, err error) {
+
+	md, _ := runtime.ServerMetadataFromContext(ctx)
+	authen := md.HeaderMD.Get(wwwAuthneticate)
+
+	statusError := status.Convert(err)
+
+	// Basic-Auth failure
+	if statusError.Code() == codes.Unauthenticated && len(authen) > 0 {
```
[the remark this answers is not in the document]
Mohammed90: Currently, these are all the places where the error code `codes.Unauthenticated` is used:

- grpc/internal/errors/errors.go: if statusError.Code() == codes.Unauthenticated && len(authen) > 0 {
- grpc/internal/errors/errors.go: if statusError.Code() == codes.Unauthenticated {
- grpc/private/private_server.go:         return ctx, grpc.Errorf(codes.Unauthenticated, "missing context metadata")
- grpc/private/private_server.go:         return ctx, status.Error(codes.Unauthenticated, `invalid base64 in header`)
- grpc/private/private_server.go:         return ctx, status.Error(codes.Unauthenticated, `invalid basic auth format`)
- grpc/private/private_server.go:         return ctx, status.Error(codes.Unauthenticated, "invalid user or password")
- grpc/public/public_server.go:           return nil, status.Error(codes.Unauthenticated, "account not found")
- grpc/public/public_server.go:                   return nil, status.Error(codes.Unauthenticated, "account")
- grpc/public/public_server_test.go:                              assert.Equal(t, codes.Unauthenticated, st.Code())
- grpc/public/public_server_test.go:                      assert.Equal(t, codes.Unauthenticated, errorStatus.Code())

The first 2 are in this file. The second set is from private_server and all correspond to basic-auth. The third set is from public_server, where there isn't basic-auth anyways, and returned from `RefreshSession` and `ChangePassword`. The last set is from tests.

The only possible scenario I can see where the program branches into this path without it actually being a basic-auth failure is when: an unauthenticated user tries to refresh a session or change password while also including the authentication header. If we're also handling this scenario, I think I have to signal it by sending a flag via `context.Context`. Do you have another suggestion?

grpc/internal/gateway/gateway.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	return nil
+}
+
+// CookieSetter extracts the session cookie from metadata and assigns it to a cookie. If the session
+// value is an empty string, then the cookie is marked to be removed.
+func CookieSetter(cfg *app.Config) GatewayResponseMiddleware {
+	return func(ctx context.Context, response http.ResponseWriter, m proto.Message) error {
+		switch m.(type) {
+		case *authnpb.LogoutResponse, *authnpb.SignupResponseEnvelope, *authnpb.LoginResponseEnvelope:
```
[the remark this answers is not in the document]
Mohammed90: Hmm. I also hate that we have to maintain this list manually, but couldn't come up with a better programmatic way. Any ideas?

grpc/internal/gateway/gateway.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+			jsonMap := make(map[string]interface{}, len(r.Form))
+			for k, v := range r.Form {
+				if len(v) > 0 {
+					jsonMap[k] = v[0]
+				}
+			}
+			jsonBody, err := json.Marshal(jsonMap)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
```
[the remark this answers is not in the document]
Mohammed90: I've added the early return. How do you like the log line?

grpc/private/gateway.go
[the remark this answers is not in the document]
cainlevy: How do you feel about that setup? I backed my way into it because I wanted to maintain backwards compatibility during 1.x while still supporting a reduced access public port. I'm currently leaning towards simplifying in 2.x.

Meanwhile, the gRPC server doesn't have any history to maintain so it can make its own choice here.

grpc/internal/gateway/gateway.go
```diff
@@ -70,13 +70,18 @@ func CookieAnnotator(app *app.App) func(ctx context.Context, req *http.Request)
...
 // FormWrapper takes form values from application/x-www-form-urlencoded and converts it to JSON
-// Workaround from: https://github.com/grpc-ecosystem/grpc-gateway/issues/7#issuecomment-358569373
+// Workaround adapted from: https://github.com/grpc-ecosystem/grpc-gateway/issues/7#issuecomment-358569373
 func FormWrapper(mux http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if strings.ToLower(strings.Split(r.Header.Get("Content-Type"), ";")[0]) == "application/x-www-form-urlencoded" {
 			if err := r.ParseForm(); err != nil {
+				// deliberately omit including the form from
+				// logging to avoid leaking sensetive data
+				log.WithError(err).WithFields(log.Fields{
```
cainlevy: i'm not usually inclined to log client faults, unless it's a situation that can be very difficult to debug from the client (e.g. CORS failures). what're your thoughts?

grpc/internal/gateway/gateway.go
```diff
@@ -70,13 +70,18 @@ func CookieAnnotator(app *app.App) func(ctx context.Context, req *http.Request)
...
+					"url":    r.URL,
+					"method": r.Method,
+				}).Warn("error parsing form")
 				http.Error(w, err.Error(), http.StatusBadRequest)
```
cainlevy: come to think of it: we should probably return an empty body rather than internal error strings.

grpc/internal/gateway/gateway.go
```diff
@@ -87,7 +92,14 @@ func FormWrapper(mux http.Handler) http.Handler {
 			}
 			jsonBody, err := json.Marshal(jsonMap)
 			if err != nil {
+				// deliberately omit including the form from
+				// logging to avoid leaking sensetive data
+				log.WithError(err).WithFields(log.Fields{
+					"url":    r.URL,
+					"method": r.Method,
+				}).Warn("error encoding form into JSON")
 				http.Error(w, err.Error(), http.StatusBadRequest)
```
cainlevy: let's hide our internal error string here as well

grpc/authn.proto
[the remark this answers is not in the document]
cainlevy: Haha, so really I'm just looking at my own choices in a mirror.

grpc/server/server.go
```diff
@@ -0,0 +1,113 @@
...
+	grpc_prometheus.EnableHandlingTimeHistogram()
+
+	if app.Config.PublicPort != 0 {
+		tcpl, err := net.Listen("tcp", fmt.Sprintf(":%d", app.Config.PublicPort))
+		if err != nil {
+			log.Fatalf("failed to listen: %v", err)
+		}
+
+		tcpm := cmux.New(tcpl)
+		grpcl := tcpm.MatchWithWriters(cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"))
```
cainlevy: was looking at the [docs](https://godoc.org/github.com/soheilhy/cmux#HTTP2MatchHeaderFieldSendSettings) and noticed:

> Prefer HTTP2HeaderField over this one, if the client does not block on receiving a SETTING frame.

thoughts?

grpc/server/server.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	tcpl, err := net.Listen("tcp", fmt.Sprintf(":%d", app.Config.ServerPort))
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
+
+	tcpm := cmux.New(tcpl)
+	grpcl := tcpm.MatchWithWriters(cmux.HTTP2MatchHeaderFieldSendSettings("content-type", "application/grpc"))
+	httpNoTLSL := tcpm.Match(cmux.HTTP1Fast())
```
cainlevy: leftover from the TLS logic:

```suggestion
	httpl := tcpm.Match(cmux.HTTP1Fast())
```

grpc/server/server.go
Mohammed90: Some clients block until they receive the SETTING frame. They cite the Java client as one of those clients in the [README.](https://github.com/soheilhy/cmux#limitations) The Go client also [has the same behavior.](https://github.com/soheilhy/cmux/issues/64)

grpc/internal/gateway/gateway.go
Mohammed90: My thought process was to be able to detect clients who are constantly submitting malformed input trying to find a hole. However, in retrospect, the logging middleware will show 400 status response anyways along with the path. Voting to remove?

grpc/private/gateway.go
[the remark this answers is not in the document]
Mohammed90: I think it's fine. To me it means I can have my public users and back-office users both exist on the same dataset, and what they're able to do is ruled by another authz microservice. It will also allow Keratin AuthN to be reused for internal apps (without having to go through the public IP address) without having to run separate instances.

grpc/internal/gateway/gateway.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+		return metadata.MD{
+			app.Config.SessionCookieName: []string{cookie.Value},
+		}
+	}
+}
+
+// FormWrapper takes form values from application/x-www-form-urlencoded and converts it to JSON
+// Workaround adapted from: https://github.com/grpc-ecosystem/grpc-gateway/issues/7#issuecomment-358569373
+func FormWrapper(mux http.Handler) http.Handler {
```
Mohammed90: As I'm writing more tests, I found that this is imperfect at converting `application/x-www-form-urlencoded` to JSON. Due to `url.Values` storing slices of strings, then the value will always be string, and it'll fail if we're expecting anything other than string. I'll be brainstorming a better convertor.

There's [this](https://github.com/grpc-ecosystem/grpc-gateway/pull/548), but it's not merged, and needs to be evaluated. Basically, we need an adapter that takes in form values, and converts the value from `string` to the appropriate type.

cainlevy: Ahh, right. The REST app normally manages type conversion in the handlers as part of the HTTP decoding process. Maybe we can encode simplifying assumptions to make it work for our limited needs? We mostly need to identify integers and booleans.

* integers: any param named `id`
* booleans: only bool currently is the `locked` param for account imports

grpc/internal/gateway/gateway.go
cainlevy: yeah, voting to remove

grpc/private/gateway.go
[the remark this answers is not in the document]
cainlevy: But do you think that the private port should bother to continue handling public endpoints? Currently the split setup looks like:

* public port -- handles public endpoints
* private port -- handles private _and public_ endpoints

The only reason the private port works that way currently is because I didn't want to introduce a breaking change in 1.x for single-port deployments. In 2.x I may switch to enforcing the two port setup.

Meanwhile, the gRPC server isn't burdened by existing deployments. It could avoid mixing public routes into the private gateway from the beginning and require a two-port configuration. Do you think there's an argument for waiting until 2.x?

grpc/private/private_server.go
```diff
@@ -0,0 +1,110 @@
...
+	s := strings.IndexByte(cs, ':')
+	if s < 0 {
+		header := metadata.Pairs("WWW-authenticate", `Basic realm="Private AuthN Realm"`)
+		grpc.SendHeader(ctx, header)
+		return ctx, status.Error(codes.Unauthenticated, `invalid basic auth format`)
+	}
+
+	user, password := cs[:s], cs[s+1:]
+	if !matcher(user, password) {
+		err := grpc.SetHeader(ctx, metadata.Pairs("www-authenticate", `Basic realm="Private AuthN Realm"`))
```
cainlevy: should this be:

```suggestion
		err := grpc.SetHeader(ctx, metadata.Pairs("WWW-authenticate", `Basic realm="Private AuthN Realm"`))
```

grpc/private/private_server.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
```
cainlevy: why do we check `err` this time but not when calling `grpc.SendHeader` earlier?

grpc/private/private_server.go
```diff
@@ -0,0 +1,110 @@
...
+			log.Errorf("error setting header: %s", err)
+		}
+		ctx = metadata.AppendToOutgoingContext(ctx, "WWW-Authenticate", `Basic realm="Private AuthN Realm"`)
```
cainlevy: why adjust context this time, but not in earlier return conditions?

grpc/public/password_reset.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	// run in the background so that a timing attack can't enumerate usernames
+	go func() {
+		err := services.PasswordResetSender(s.app.Config, account)
+		if err != nil {
+			s.app.Reporter.ReportError(err)
```
cainlevy: should we now prefer `ReportGRPCError` in all grpc handlers?

* `grpc/private/unsecured.go`
* `grpc/public/password_reset.go`
* `grpc/public/passwordless.go`
* `grpc/public/public_server.go`

grpc/public/signup.go
```diff
@@ -0,0 +1,77 @@
...
+		}, nil
+	}
+
+	br := &errdetails.BadRequest{}
+	br.FieldViolations = append(br.FieldViolations, &errdetails.BadRequest_FieldViolation{
+		Field:       "username",
+		Description: services.ErrTaken,
+	})
+
+	statusError := status.New(codes.FailedPrecondition, services.FieldErrors{{Field: "username", Message: services.ErrTaken}}.Error())
```
cainlevy: worth going through `errors.ToStatusErrorWithDetails` here for consistency and de-duplication?

grpc/public/public_server.go
```diff
@@ -0,0 +1,211 @@
...
+func (s publicServer) Login(ctx context.Context, req *authnpb.LoginRequest) (*authnpb.LoginResponseEnvelope, error) {
+	account, err := services.CredentialsVerifier(
+		s.app.AccountStore,
+		s.app.Config,
+		req.GetUsername(),
+		req.GetPassword(),
+	)
+	if err != nil {
+		if fe, ok := err.(services.FieldErrors); ok {
+			br := &errdetails.BadRequest{}
```
cainlevy: prefer `errors.ToStatusErrorWithDetails`?

grpc/public/public_server.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// check for valid session with live token
+	accountID := meta.GetSessionAccountID(ctx)
+	if accountID == 0 {
+		return nil, status.Error(codes.Unauthenticated, "account not found")
+	}
+
+	identityToken, err := services.SessionRefresher(
+		s.app.RefreshTokenStore, s.app.KeyStore, s.app.Actives, s.app.Config, s.app.Reporter,
+		meta.GetSession(ctx), accountID, &s.app.Config.ApplicationDomains[0],
```
cainlevy: there are a few places where the REST server uses `route.MatchedDomain` to reflect the domain validated from an `Origin` or `Referer` header back as the `aud` claim. this is intended to support cross-domain logins, with an access token for each domain.

is it possible to support that with the gRPC server? it could matter to JavaScript clients that connect to gRPC through the REST interface.

grpc/private/gateway.go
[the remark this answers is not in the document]
cainlevy: Found my own answer: backwards compatibility remains valuable to the gRPC server so that it can be a drop-in replacement (albeit experimental). Happy to defer simplifying all of this until 2.x.

glerchundi:
I came across to this PR when digging into a previous issue. One that you referenced here and I thought this would be interesting for you to see how we're solving public vs internal services problem.

Hope it helps: https://github.com/uber/prototool/issues/469#issuecomment-509284046.

grpc/public/public_server.go
Mohammed90: I probably can pull something together using the [`:authority`](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md) header/metadata to emulate this. Let me look into it and experiment to figure something out.

grpc/private/routing.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/keratin/authn-server/server/handlers"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+/*
+	Reference: github.com/keratin/authn-server/server/private_routes.go
+*/
+
+// RegisterRoutes registers gmux as the handler for the private routes on router
+func RegisterRoutes(router *mux.Router, app *app.App, gmux *runtime.ServeMux) {
```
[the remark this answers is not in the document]
Mohammed90: Would it be better if I move both the public & private routes registration to the internal package? This way we won't have the awkwardness of public route registration being exported but not the analogous private routes.

grpc/public/routing_test.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func TestRegisterRoutes(t *testing.T) {
+	testOAuthProvider := oauth.NewProvider(&oauth2.Config{}, func(t *oauth2.Token) (*oauth.UserInfo, error) {
+		return &oauth.UserInfo{
+			ID:    "test@example.com",
+			Email: "test@example.com",
+		}, nil
+	})
+	t.Run("App with SignupEnabled & AppPasswordResetURL enabled has all routes registered", func(t *testing.T) {
```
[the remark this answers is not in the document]
Mohammed90: My intent was to treat `pathMethods` as the source of truth to compare the registered paths against it. To be honest, I didn't like it much when I wrote that down, but put it to get started writing tests for the routing registration until a better way is figured out. Can you think of something? I know it looks gross 😅

grpc/public/password_reset.go
Mohammed90: Hmm... the problem is we don't have `*grpc.UnaryServerInfo` in here. This information is only available at the interceptor level. Should I pass it via `context`? Or, and this doesn't feel clean, panic and let the panic-recovery interceptor handle it which can inject `*grpc.UnaryServerInfo`?

grpc/private/private_server.go
Mohammed90: This is a mistake. [`AppendToOutgoingContext` is for client-side.](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#sending-and-receiving-metadata---client-side)

Mohammed90: After looking into this, we don't need to set any headers in this case. I used the SetHeader to signal to the gRPC-Gateway error handler that we've encountered Basic-Auth error. But the HTTP Basic-Auth isn't validated on this level. It's already checked by the HTTP middleware before being passed-through to the gRPC handlers. I've removed all the {Set,Send}Header stuff. Returning the error suffices.

grpc/public/routing_test.go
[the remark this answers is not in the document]
cainlevy: yeah the `pathMethods` collection is awkward but pragmatic.

i was mostly observing that the test name implies that it's going to verify all members of `pathMethods` are present in the router, but instead it verifies that all members of the router are present in `pathMethods`.

let's update the test name _or_ the test logic

grpc/public/password_reset.go
cainlevy: 👌 i understand. i'm happy to defer this decision until we have real-world experience that motivates extra complexity.

Mohammed90: Actually, I've solved it locally but haven't pushed the changes yet. I opted to inject `*grpc.UnaryServerInfo` via interceptor through `context.Context`, and retrieve it from the context when needed.

Mohammed90:
Do you think it'd be better to split the proto files and the generated files into their own repo? This way, those using it as gRPC clients don't end up pulling the entire AuthN repo. I haven't explored this path with experimentation, just thinking.

silasdavis:
> Do you think it'd be better to split the proto files and the generated files into their own repo? This way, those using it as gRPC clients don't end up pulling the entire AuthN repo. I haven't explored this path with experimentation, just thinking.

I'd much rather not personally. The protobuf files change with the repo and it feels like this just adds busy work. Clients can pull individual files from github raw via permalink if they like. But cloning authn is hardly a big deal if you are depending on it anyway.

Mohammed90:
It seems possible to get rid of the envelopes by wrapping the outgoing message with the gateway's `WithForwardResponseOption`. That'd make for cleaner proto messages. I'll fiddle with it and see how it goes.

grpc/authn-private.proto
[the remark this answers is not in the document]
Mohammed90: I just had a thought that could make this better. What if I switch the names around, i.e.:
```
message GetAccountResponse {
    EnvelopedGetAccountResponse result = 1;
}
```

This way if decide to change the enveloped data, the name of the message remains untouched, still descriptive of the content, and backwards compatible. Thoughts? If agreed, I can do this along with the merge-conflict resolution.

dhillondeep:
What is the status of this PR?

Mohammed90:
> What is the status of this PR?

Although I worked on it, both Lance and I have talked about how the size of the change makes worrying, thus a bit hesitant to accept it. The size of it wasn't predicted as the approach was largely exploratory, i.e. we knew we'll be using gRPC-gateway (because it harmonizes gRPC API with REST API) but we can't foresee full architecture of the change because neither of us have used gRPC-gateway before.

dhillondeep:
> Although I worked on it, both Lance and I have talked about how the size of the change makes worrying, thus a bit hesitant to accept it. The size of it wasn't predicted as the approach was largely exploratory, i.e. we knew we'll be using gRPC-gateway (because it harmonizes gRPC API with REST API) but we can't foresee full architecture of the change because neither of us have used gRPC-gateway before.

I am working on grpc microservices and was thinking of using this project. I agree there will be substantial changes to the way this project is structured. Is it possible to open this conversation again because grpc support will help many people like me trying to use this project. GRPC has grown a lot in recent time and with proper discussion and planning, I think we could do this! I can help with this as well but I would like to ask what you all think

cainlevy:
I agree with @Mohammed90's description. There were a couple of unresolved design questions, but mostly it was a lot of work to create and review. I also expected it to be a maintenance challenge for new contributors.

A few paths forward might look like:

* a dedicated gRPC fork, using go imports to full advantage
* some reduced scope of work
* some new approach to gRPC & Rest hybrid

dhillondeep:
You are right! In my opinion the safest option would be to have this work carried over in a fork and both projects can sync in terms of functionality in the meantime. If in future the fork gets mature enough that it can handle REST then a possible switch could be possible. This also allows grpc project to develop incrementally and not have to rely on breaking things.

I have forked the project to https://github.com/uptimize/authn-grpc and would like to start with this!

cainlevy:
I'm inclined to close this PR and redirect attention to your fork. What are your opinions @Mohammed90 ?

Mohammed90:
> I'm inclined to close this PR and redirect attention to your fork. What are your opinions @Mohammed90 ?

Sure, go ahead. I wasn't sure whether you wanted to keep it for discussion or not, and that's why I kept it open.

PR #93, opened 2019-04-08
[Client] Allow use of custom HTTP client in route/client.go

silasdavis:
I was writing a Go client wrapper around `Signup` and some other methods, looking at the authn-lib lib/route client I thought I may as well try and reuse that. However I wanted to be able to set up HTTP to user mutual TLS so needed a way to provide the HTTP client. This PR adds the ability to do that.

Incidentally, would it make sense for the go client library (https://github.com/keratin/authn-go) just to be part of authn-server? It seems like there is more comprehensive functionality here than there anyway, along with convenient wrapper types etc.

### DIFF
--- a/lib/route/client.go
+++ b/lib/route/client.go
@@ -15,6 +15,7 @@ type modder func(*http.Request) *http.Request
 type Client struct {
 	BaseURL   string
 	Modifiers []modder
+	*http.Client
 }
 
 const (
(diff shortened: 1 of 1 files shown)


silasdavis:
> Maybe the other way? This client could move to authn-go and be imported into authn-server. I'm reluctant to privilege the authn-go client by making it the one client in a shared repo.

I see where you are coming from regarding not privileging a particular language, but in a way the Go client _is_ special, it's the reference client for testing. Do you really want to have to flip flop between the two repos and coordinate dependent releases to get a version that will work with tests? Perhaps it is more stable than that, but given that authn-go doesn't support most of the authn API right now, perhaps it would be better to incubate it within the main repo, reduce the overhead of managing two repos and releases processes, and spin it out when things are more stable and you have more hands on deck.

cainlevy:
Yeah, I think it'd be fine. The tests aren't driving changes anymore.

On the other hand, the AuthN backend clients may convert to gRPC and the server may be the only place that still needs a REST client for tests.

Either way, I think we can solve for sharing HTTP client logic without moving the golang client. My preference is to either keep repos equally separated (while taking advantage of Go's robust import and module system) or equally consolidated (monorepo style).

Issue #94, opened 2019-04-08
Using authn-server as a library?


silasdavis:
I note that the intended mode of operation for authn-server is for it to be run as an isolated service exposed to your front end, e.g. https://github.com/keratin/authn-server/issues/79.

I was initially sold on the principle of isolation for the authentication service but I have an number of requirements that make me wonder if I would be better off integrating authn-server as a library into a Go service of my own. In particular I need to:

1. To synchronize with another database (a blockchain) on user signup (and possible on other account changes)
2. Expose a consent app to Ory Hydra to act as an OpenID connect provider
3. Ideally talk to other services over GRPC (I know that is in hand)

I think I could work around all of these with various separate services, but it feels like I may end up doing a lot of busywork to write a Go app to deal with these. 1 might be fundamentally better solved by dealing with things atomically in a single request scope.

In any case I will be building an account management service for multiple backend services (privacy settings, email sending, organisation admin etc) to handle this across our platform so I'm going to end up with two quite tightly coupled services to operate, where my account service ends up proxying at least the signup call to authn-server.

My question is where do you stand on using authn as a library? Is it something that totally subverts the design you had in mind, or is it just that I would need to be prepared for breakages in internal APIs? They way things are structured now seems open to extension - I think I could live with them not being stable.

Perhaps there is scope for some kind of synchronisation/callback interface that maintains the boundary but allows for some tightly coupled logic to run with authentication/signup. It might work a bit like an extension to `WebhookSender`, perhaps implemented as a GRPC stream, or if building in Go could work by running authn-server within a Go wrapper and subscribing to events.

cainlevy:
Let me turn this back to you: what's the value proposition of authn-server as a library?

I've seen a lot of authentication libraries. I've seen a fair number of third-party authentication services. My approach to authn-server has been to explore the middle ground, gaining isolation without giving over account management to another brand.

This means I've made choices that optimize authn-server for a particular setup. If you decided to seek a different setup, then I'd be curious to know what you valued in authn vs some other authentication library.

> Perhaps there is scope for some kind of synchronisation/callback interface that maintains the boundary but allows for some tightly coupled logic to run with authentication/signup.

I find this direction more interesting. The boundary that authn-server creates has worked fairly well for me (after dealing with atomic signup).

[...]

PR #98, opened 2019-05-13
Use sqlx.Ext interface rather than *sqlx.DB in all account_store.go

silasdavis:
This if the first of my PRs in relation to https://github.com/keratin/authn-server/issues/94. Hopefully it meets your criterion of:

> I'm happy to accept improvements that make AuthN easier to use as a library whenever those improvements are at least net neutral for AuthN as a microservice. 

The interface allows static access to the same methods but allows the
use of a DB or a transaction which faciliates easier extension of
authn-server when used as a library.

For example if you wish to create a user profile or other resources
atomically you may pass a sqlx transaction with sqlx.Beginx

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1,6 +1,7 @@
 package app
 
 import (
+	"github.com/jmoiron/sqlx"
 	"os"
 
 	"github.com/go-redis/redis"
@@ -16,6 +17,7 @@ import (
 type pinger func() bool
 
 type App struct {
+	DB                *sqlx.DB
 	DbCheck           pinger
 	RedisCheck        pinger
 	Config            *Config
@@ -97,6 +99,8 @@ func NewApp(cfg *Config) (*App, error) {
 	}
 
 	return &App{
+		// Provide access to root DB - useful when extending AccountStore functionality
+		DB:                db,
 		DbCheck:           func() bool { return db.Ping() == nil },
 		RedisCheck:        func() bool { return redis != nil && redis.Ping().Err() == nil },
 		Config:            cfg,
--- a/app/data/account_store.go
+++ b/app/data/account_store.go
@@ -27,14 +27,14 @@ type AccountStore interface {
 	SetLastLogin(id int) (bool, error)
 }
 
-func NewAccountStore(db *sqlx.DB) (AccountStore, error) {
+func NewAccountStore(db sqlx.Ext) (AccountStore, error) {
 	switch db.DriverName() {
 	case "sqlite3":
-		return &sqlite3.AccountStore{DB: db}, nil
+		return &sqlite3.AccountStore{Ext: db}, nil
 	case "mysql":
-		return &mysql.AccountStore{DB: db}, nil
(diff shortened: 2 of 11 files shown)


silasdavis:
Come to think of it... do we really need an `AccountStore` at all? I think it might be cleaner just to have each method take a `sqlx.Ext` as the first argument and make it clearer there is no in-memory state involved here and the methods can be static.

silasdavis:
Hmm on reflection I think we probably need to leave the `AccountStore` types alone. What I really want is static access to the `AccountStore` methods with a first argument of `sqlx.Ext`. This is easily accomplished by splitting them out though there is a certain amount of boiler plate in implementing each of the interface methods against the `AccountStores` however... this might be an good opportunity to deduplicate some of the query code where possible.

Let me know if you would consider this and I'll have a go and implementing some shared static exported methods between the various store implementations.

silasdavis:
For a bit of context, here is my extension to `AccountCreator` in my project:

```go
func AccountCreator(app *app.App, username, password string,
	profile *models.Profile) (*models.Account, error) {

	err := profile.Validate()
	if err != nil {
		return nil, err
	}

	// Create account and profile in single shot
	tx, err := app.DB.Beginx()
	if err != nil {
		return nil, err
	}
	// Just silently err on success
	defer tx.Rollback()

	// Use TX in store
	store := &postgres.AccountStore{Ext: tx}

	keratinAccount, err := services.AccountCreator(store, app.Config, username, password)
	if err != nil {
		return nil, err
	}

	accountID := keratinAccount.ID

	profile.AccountID = accountID

	profile, err = data.CreateProfile(tx, profile)
	if err != nil {
		return nil, err
	}

	err = tx.Commit()
	if err != nil {
		return nil, err
	}

	return &models.Account{
		Account: *keratinAccount,
		Profile: *profile,
	}, nil
}
```

Incidentally, we were thinking we could probably open source this - it may be a useful example for others wanting to do something similar (i.e. SSO/centralised platform account with authn as a library). Though not sure how reusable it is at the moment.

cainlevy:
Creating a new `AccountStore` for each transaction seems effective. Do you see problems with it?

silasdavis:
@cainlevy yeah its fine actually.. the thing I really wanted to avoid was this:

```go
	// TODO: replace this hack with something we can use that is palatable enough for upstream to merge (i.e. exposing sqlx.DB whilst stilling getting hold of methods)
	as, ok := keratinApp.AccountStore.(*postgres.AccountStore)
	if !ok {
		return nil, fmt.Errorf("%s: could not extract postgres AccountStore from the Keratin App", errHeader)
	}
	db, ok := as.Ext.(*sqlx.DB)
	if !ok {
		return nil, fmt.Errorf("%s: could not extract DB from AccountStore", errHeader)
	}
```

But if you are happy with exposing the `sqlx.DB` off `App` as I have added (but squashed) in this PR then I can just do:

```
	db := keratinApp.DB
```

silasdavis:
Rebased.

PR #99, opened 2019-05-13
Fix zxcvbn divergence

silasdavis:
While testing my sign up form I noticed that some passwords meeting my strength criteria in JS were failing remotely.

It looks like the zxcvbn library currently used by authn is not conformant: https://github.com/nbutton23/zxcvbn-go/issues/20 

This one looks better, and at the very least passes my single test case: https://github.com/trustelem/zxcvbn and so this PR switches us over to that library.

They seem to have a reasonably good testing regime so hopefully it can be trusted. Pretty sure it is an improvement in any case.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .env
 vendor/
 dist/
+.idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 * panic while evaluating some utf8 password characters
+* zxcvbn library we use exhibited some deviation from standard (see: https://github.com/nbutton23/zxcvbn-go/issues/20) so switched to https://github.com/trustelem/zxcvbn [#99]
 
 ## 1.5.0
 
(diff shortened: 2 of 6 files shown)


app/services/validations_test.go
```diff
@@ -1,14 +1,31 @@
-package services_test
+package services
```
silasdavis: I realise I am diverging from the house style here. I didn't think the black box tests were providing much value here, but if you like I can alternatively:

1. Introduce an internal test with access to the `passwordValidator` function
2. Export `passwordValidator`
3. Import our zxcvbn library and test it directly
4. Just choose to believe this library's own tests and assertions and omit the test altogether

cainlevy: I'm happy with 4 because we can't do better than:

> all unit tests from the upstream library have been ported (and even more tests have been added) to ensure that this holds.

from https://github.com/trustelem/zxcvbn/blob/master/README.md

silasdavis:
Done thanks for review

PR #105, opened 2019-05-31
Allow same-origin requests

### LINKED ISSUE #67: CORS issue with same-origin deployment
Hi,

If running authn-server in k8s behind an ingress which has the following path based routing:
/auth -> authn
/         -> website

When a user logs into the site a CORS error will be returned as since the domain is the same for the post the browser will not send an `Origin` header or attempt pre-flight.

I read the previous comment about disabling CORS and agree that it should be in place however I am wondering if there is a fix for same domain hosted applications.

silasdavis:
By infering domain from Referer header iff Origin header is missing.

Some browser (e.g. Firefox) do not send Origin header for same origin requests. Since we implicitly rely on browser to add the header for any cross-origins requests we ought to similarly be able to infer that it is _not_ a cross-origin request when the Origin header is missing. This encodes that assumption and as suggested in #67 uses the Referer header to observe the same origin in question. We fail if the Referer header is missing and we only ever consider the Referer header if the Origin header is missing.

Closes #67.

### DIFF
--- a/lib/route/origin_security_test.go
+++ b/lib/route/origin_security_test.go
@@ -11,35 +11,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestOriginSecurity(t *testing.T) {
-	readBody := func(res *http.Response) []byte {
-		body, err := ioutil.ReadAll(res.Body)
-		require.NoError(t, err)
-		res.Body.Close()
-		return body
-	}
+const successBody = "success"
 
+func TestOriginSecurity(t *testing.T) {
 	testCases := []struct {
-		domain  string
-		origin  string
-		success bool
+		domain      string
+		goodOrigins []string
+		badOrigins  []string
 	}{
-		{"example.com", "http://example.com", true},
-		{"example.com", "http://example.com:3000", true},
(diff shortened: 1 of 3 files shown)


lib/route/origin_security_test.go
```diff
@@ -49,17 +37,57 @@ func TestOriginSecurity(t *testing.T) {
...
-			res, err := http.DefaultClient.Do(req)
-			require.NoError(t, err)
+			for _, goodOrigin := range tc.goodOrigins {
+				// Cross-site case
+				assertOriginAccepted(t, server, goodOrigin, goodOrigin)
+				assertOriginAccepted(t, server, goodOrigin, "")
+				// Same-origin case
+				assertOriginAccepted(t, server, "", goodOrigin)
+				// Undefined - referer and origin missing
+				assertOriginDenied(t, server, "", "")
```
cainlevy: pull this one out of the loop?

silasdavis: ah yes, the '(bad, bad)' one can come out too...

silasdavis:
Pulled out some assertions from loop and added to changelog

PR #106, opened 2019-05-31
add PUT support to the REST API

### LINKED ISSUE #104: No Routes Registered for PUT
As I'm writing the integration test for #83, I ran into PUT calls returning HTTP 405, which seems reasonable because the code never registers routes for PUT and only does so for PATCH. However, the [API documentation](https://github.com/keratin/authn-server/blob/7f82ad87d4ff3439992e056a1219cda4a18b72c6/docs/api.md) specifies that every every route accepting PATCH also accepts PUT.

We can:
- Removing references of PUT in the documentation
- Add support for PUT

Either of those solutions wouldn't be a breaking change. Which one is preferred? 

Here are the relevant files segments.

https://github.com/keratin/authn-server/blob/7f82ad87d4ff3439992e056a1219cda4a18b72c6/server/private_routes.go#L14-L58

https://github.com/keratin/authn-server/blob/7f82ad87d4ff3439992e056a1219cda4a18b72c6/lib/route/route.go#L15-L40

Mohammed90:
Fixes #104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 * Log when rejecting a request for a missing or invalid Origin header [#34]
+* Accept PUT HTTP calls on every endpoint accepting PATCH [#104]
 
 ### Improved
 
--- a/lib/route/route.go
+++ b/lib/route/route.go
@@ -34,6 +34,11 @@ func Patch(tpl string) *Route {
 	return &Route{Verb: "PATCH", Tpl: tpl}
 }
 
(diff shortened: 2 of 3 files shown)


Mohammed90:
Code surface sans tests increased -> test coverage increased.... can't explain that 🤷‍♂

silasdavis:
Ha.

All of the code you added is touched by tests since test server ->  router -> private routes so the proportion has increased.

LGTM

Mohammed90:
> 🤔 seems like a fine solution. would you add an entry to the changelog?

Done

PR #115, opened 2019-07-11
Use an explicit logrus FieldLogger rather than global logger.

silasdavis:
It is convenient to use a global logger, but it makes it hard to provide coherent logging if you are a dependee using keratin as a library. Moreover it is another source of global state that can cause unwanted effects.

This PR passes loggers as an explicit dependency. I have added to the logger to the `Config` object. At first I wanted to add it to `App` because it doesn't really seem like config. However I added to `Config` in the end because it seems to fit better and:

- `ErrorReporter` doesn't seem like config but is initialised in `Config`. Logger doesn't seem like config either (maybe two wrongs don't make a right)
- If I add it to `App` I would either need to make some functions take `App` that take `Config` (making it even fuzzier what they actually need) or add `logger` as an explicit last argument (I tend to follow this approach in my own code, but it is more susceptible to the complaint that 'loggers get everywhere' - they get where they are used in any case)

I also added a hook that reports logrus errors to our `ErrorReporter`. This is unrelated to the above so I can drop if you like, but I thought it made sense and logrus provides hooks out of the box so I thought why not. If you don't want to keep that I can add a test and make the coverage go up instead of down.

### DIFF
--- a/lib/route/origin_security.go
+++ b/lib/route/origin_security.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"net/http"
 
-	log "github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus"
 )
 
 type matchedDomainKey int
@@ -15,12 +15,12 @@ type matchedDomainKey int
 //
 // OriginSecurity will store the matching domain in the http.Request's Context. Use MatchedDomain
 // to retrieve the value in later logic.
-func OriginSecurity(domains []Domain) SecurityHandler {
+func OriginSecurity(domains []Domain, logger logrus.FieldLogger) SecurityHandler {
 	var validDomains []string
 	for _, d := range domains {
 		validDomains = append(validDomains, d.String())
 	}
-	logger := log.WithFields(log.Fields{"validDomains": validDomains})
+	logger = logger.WithField("validDomains", validDomains)
 
 	return func(h http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -37,7 +37,7 @@ func OriginSecurity(domains []Domain) SecurityHandler {
 			if len(origin) == 0 {
 				logger.Debug("Could not infer request origin since Origin and Referer headers are both missing")
 			} else {
-				logger.WithFields(log.Fields{"origin": origin}).Debug("Request origin is invalid")
+				logger.WithField("origin", origin).Debug("Request origin is invalid")
 			}
 			w.WriteHeader(http.StatusForbidden)
 			w.Write([]byte("Origin is not a trusted host."))
--- a/app/app.go
+++ b/app/app.go
@@ -27,13 +27,10 @@ type App struct {
 	Actives           data.Actives
 	Reporter          ops.ErrorReporter
 	OauthProviders    map[string]oauth.Provider
+	Logger            logrus.FieldLogger
 }
 
 func NewApp(cfg *Config) (*App, error) {
-	logrus.SetFormatter(&logrus.JSONFormatter{})
-	logrus.SetLevel(logrus.DebugLevel)
-	logrus.SetOutput(os.Stdout)
-
 	errorReporter, err := ops.NewErrorReporter(cfg.ErrorReporterCredentials, cfg.ErrorReporterType)
 
 	db, err := data.NewDB(cfg.DatabaseURL)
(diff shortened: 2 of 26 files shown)


lib/route/origin_security.go
```diff
@@ -34,11 +31,6 @@ func OriginSecurity(domains []Domain) SecurityHandler {
 				return
 			}
 
-			if len(origin) == 0 {
-				logger.Debug("Could not infer request origin since Origin and Referer headers are both missing")
```
silasdavis: We could pass a logger through here, but on reflection these seem like debug printfs and it would probably be better to test these conditions rather than log them.

app/config.go
```diff
@@ -62,6 +62,7 @@ type Config struct {
 	DailyActivesRetention       int
 	WeeklyActivesRetention      int
 	ErrorReporter               ops.ErrorReporter
+	Logger                      logrus.FieldLogger
```
silasdavis: This is the main interface provided by logrus - so its possible to provide your own implementation as a dependee

app/config.go
```diff
@@ -70,6 +71,20 @@ type Config struct {
 	FacebookOauthCredentials    *oauth.Credentials
 }
 
+// Lazy initialize Logger for code using partially initialized Config
+func (c *Config) GetLogger() logrus.FieldLogger {
```
silasdavis: This does make Logger a bit of special case (perhaps it is), but hs the least impact on tests that regularly pass semi-initialized config. Since it waits until a logger is needed it also gives the best change ErrorReporter is initialised.

Having it on Config will play nicely with any future desire to expose logging options in `configure` later (like which formatter to use).

ops/error_reporter.go
```diff
@@ -34,3 +36,38 @@ func PanicHandler(r ErrorReporter, next http.Handler) http.Handler {
 		next.ServeHTTP(w, req)
 	})
 }
+
+// Not sure why logrus doesn't just provide this...
+type logrusHook struct {
```
silasdavis: If you want to keep this I suspect you'll prefer it in a different file, I can move it and add tests.

app/config.go
```diff
@@ -70,6 +71,20 @@ type Config struct {
...
+	if c.Logger == nil {
```
cainlevy: if `GetLogger` sticks around, this could be a good case for [sync.Once](https://godoc.org/sync#Once).

ops/error_reporter.go
cainlevy: how well does this fit with the assortment of glue code that you're collecting?

i see theoretical value in relaying log errors to the configured reporter, but so far keratin/authn-server has preferred sending errors directly to the reporter without going through logs (unless the reporter is LogReporter).

cainlevy:
https://github.com/keratin/authn-server/pull/117

app/config.go
silasdavis: Yes I suppose there could be a race here. In the current code I'm fairly certain there is not since `GetLogger` gets called early in initialisation before we have spun off goroutines. I might refactor it away otherwise why not.

silasdavis:
Fine with these changes. Not sure if I'll get to it this week, but will do.

ops/error_reporter.go
silasdavis: This doesn't necessarily need to be a part of authn - I could implement it downstream.

I think reporting errors to both the logs and the error reporter makes most sense - yet it is clean to just pipe into the error reporter as you currently do. So I like automatically sending it to logger.

The case where it makes less sense of course is if the ErrorReporter is a LogReporter.

silasdavis:
Made those changes, works fine.

Worth merging #117 then I can fix up linking with reporter hook.

I've left error reporting as part of the `DefaultLogger` if you want this removed I can. I now only report to ErrorReporter if the reporter is not itself a `LogReporter`.

I also changed `LogReporter` to use logrus, which seemed more consistent.

When I was fixing things up I `goimport`d everything, so uhm I've touched a lot of files I didn't need to. In case like me you prefer goimported imports I've left those changes, I can roll them back if you want.

cainlevy:
> I also changed LogReporter to use logrus, which seemed more consistent.

✅ 

> When I was fixing things up I goimportd everything

Happy to clean up imports in a separate pull request. I'd like this changeset to remain focused for review and for posterity.

silasdavis:
Removed all goimports changes to unrelated files

PR #119, opened 2019-08-04
Update config.md

obukhov:
lib/route/domain.go:16 -`func ParseDomain` splits by colon symbol, so if the schema is provided, it is not parsed correctly.


cainlevy:
Thanks!

PR #126, opened 2019-08-17
proper loading of docker secrets

Kulak:
Main focus is on ability to pass Docker secrets properly instead of environment variables.

If Secure settings are not set, then drop to original mode of loading secrets from environment variables.

Other minor changes:

- App name and version are printed as 1st line on startup to to make sure app is starting
- Added Docker version tags
- removed routing.go from Makefile, because it does not seem to exist (could it be the reason for runtime error I reported?)

Version with these changes is on my Docker space https://cloud.docker.com/repository/docker/kulak/authn-server

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ ORG := keratin
 PROJECT := authn-server
 NAME := $(ORG)/$(PROJECT)
 VERSION := 1.6.0
-MAIN := main.go routing.go
+MAIN := main.go
 
 .PHONY: clean
 clean:
@@ -43,7 +43,7 @@ dist/authn-windows64.exe: init
 # The Docker target wraps the linux/amd64 binary
 .PHONY: dist/docker
 dist/docker: dist/authn-linux64
-	docker build --tag $(NAME):latest .
+	docker build --tag $(NAME):latest --tag $(NAME):$(VERSION) .
 
 # Build all distributables
 .PHONY: dist
--- a/app/config.go
+++ b/app/config.go
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"io/ioutil"
 	"math/big"
 	"net/url"
 	"os"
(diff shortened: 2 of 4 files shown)


cainlevy:
@Kulak Thanks for contributing! You've definitely caught a bug in the Makefile, and I like the feedback of logging the server name and version earlier. Happy to merge those on sight.

I'd like to understand Docker Secrets a bit better though. The only bits I can find online suggest that it's only available with Docker Swarm, right?

Is there something else I can do to improve on ENV vars for folks who are not using Docker Swarm? I've had a notion to refactor with something like https://github.com/namsral/flag or https://github.com/peterbourgon/ff to provide flexibility between command line options, ENV vars, and a config file.

Ideally a solution would not require listing specific ENV vars that have file versions, but will treat all of them with equal safety.

Kulak:
Yes, Docker needs to be in swarm mode.  

No you don't have to use Docker in swarm mode to use secrets.  Confusing, right :)

I use docker-compose and in it I can reference secrets even docker is not in swarm mode.  

Here is example file I had for your project:

```
services:
  authn:
    container_name: authn
    image: kulak/authn-server:latest
    ports:
      # host:container
      - "8765:3000"
    environment:
      - DATABASE_URL_FILE=/run/secrets/authn_db_str
      - REDIS_URL=redis://redis:6379/0
      - AUTHN_URL=http://authn:3000
      - APP_DOMAINS=localhost
      - SECRET_KEY_BASE_FILE=/run/secrets/authn_secret_key_base
    networks:
      - home
    secrets:
      - authn_db_str
      - authn_secret_key_base

networks:
  home:
    external: true

secrets:
  authn_db_str:
    file: /projs/secrets/authn-db-str
  authn_secret_key_base:
    file: /projs/secrets/authn-secret-key-base

version: "3.2"
```

When I run this file on Docker in swarm mode, I got a warning that docker was in swarm mode.  I turned it off and warning is gone.

By the way, the way I did PR had a goal of not breaking existing configuration approach.  I don't advocate for my solution as a long term solution.

Unfortunately, I run into a more serious problem with the project than just a secrets loading.  I apologize if I am wrong.  

I found that the docker image is not usable the way it is configured.  It seems that application needs to run with migrate argument when app runs 1st time. Then application needs to run with "server" option.  Well, that's kind of not easy to accomplish with a single container.  

Unless "server" actually runs migrate automatically and I did not realize.

I would be going beyond normal docker usage patterns if I have to somehow start container in merge mode or in server mode.  So, I either miss something or docker usage has issues.

cainlevy:
>  It seems that application needs to run with migrate argument when app runs 1st time.

Yeah, that's true. I'm aware of two viable strategies:

1. Run migrate separately (manually or as part of CD pipeline)
2. Override the container's default command with `sh -c "./authn migrate && ./authn server"` to execute both

Kulak:
I can close pull request.  I just wanted to expose your project to secrets.  So, no big deal.  I think the goal is accomplished.

cainlevy:
It's appreciated! I wasn't aware of that Docker system.

Would you like to open another pull request (or rebase this one) for the other changes you caught? If it's just hassle, I can also make the commits directly.

Kulak:
I am sorry.  That commit is all I have had time for regardless this project.  Feel free to close without incorporating or accept as is.  I intend to put no additional effort here.

cainlevy:
Thanks for the contribution @Kulak. I've applied the smaller improvements directly to master and will take more time to consider how secrets are loaded.

PR #128, opened 2019-09-02
docs: fix github link (corner widget)

srenatus:
I guess it should go there; before, it would end up in a 404 page.


cainlevy:
Thanks!

PR #129, opened 2019-09-04
Add the nocgo build flag to enable a build without sqlite3

silasdavis:
This involves adding some stub files that panic if they are ever used on
a nocgo build - i.e. one excluding sqlite3 but the default build will
build with sqlite3 support using CGO.

CGO is great if you need it, but can cause various types of failures and
complexities particularly on on-standard toolcahins. As a library user I if I do not use sqlite3 I would rather not pay the prices of using CGO.

In particular at the time of writing static CGO builds with go 1.13 do
not work for me on musl-gcc or Arch Linux.

### DIFF
--- a/app/data/sqlite3/db.go
+++ b/app/data/sqlite3/db.go
@@ -1,3 +1,5 @@
+// +build !nocgo
+
 package sqlite3
 
 import (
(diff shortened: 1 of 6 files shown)


app/data/sqlite3/db.go
```diff
@@ -6,19 +8,23 @@ import (
 
 	"github.com/jmoiron/sqlx"
 	// load sqlite3 library with side effects
-	_ "github.com/mattn/go-sqlite3"
+	sq3 "github.com/mattn/go-sqlite3"
 )
 
+type Error = sq3.Error
```
silasdavis: Pushing these aliases here just means we can isolate the CGO requiring files and wont' need another  `_nocgo.go` file elsewhere

Mohammed90:
The double-negative tag of `!nocgo` is awkward. Wouldn't it be better to to switch around the logic to the tags `cgo` and `!cgo`?

silasdavis:
I'll add a makefile target.

Would we prefer `!nosqlite` as a tag?
+ more explicit about what support is being turned off (though perhaps less about why you would want to do that)
- if there is ever another cgo dependency we'll need another negated tag for that dep

cainlevy:
i prefer `nocgo`. it describes the root cause, and does not create questions about `nomysql`, `nopostgresql`, and `noredis`.

AlexCuse:
I think about this one sometimes when I think about ARM builds. Do we have any particular requirements for SQLite?  There is eg https://pkg.go.dev/modernc.org/sqlite offering a pure go driver, it might not be as performant but our SQL usage is really simplistic it might be enough.

cainlevy:
Sqlite is a good option for simple installs but is not necessary when Redis and either Postgres or MySQL are available.

PR #133, opened 2019-10-03
Implement TOTP MFA

### LINKED ISSUE #92: Add support for Multifactor Authentication (MFA / TFA)
Is this something you would consider/have considered? It seems to me like it would be an appropriate features to situate in authn-server and It's something I need.

For example, by integrating: https://github.com/pquerna/otp 

I would be willing to work on this, if you have some initial guidance then I could give a draft design / implementation proposal.

thenhawke:
Same as #132 (didn't mean to close the last PR 🖥🐒)

Fixes #92 

> Currently the TOTP cache uses the same TTL as the ACCESS_TOKEN_TTL, should I create a new environment variable for this value? That means currently they have ACCESS_TOKEN_TTL time between the GET /totp and the POST /totp to enable TOTP on their account.

> Currently the TOTP secret is saved to the database (postgres, mysql) with compat.Encrypt and the app's config.DBEncryptionKey since the secret is required in plaintext to verify. Is this ok?


thenhawke:
> handlers/get_oauth_return.go - MFA after OAuth seems unusual. haven't seen it before. you?

No, I don't think I have, I think the MFA would have to be handled by the OAuth provider in that case. (Google or Github) 

>handlers/post_password.go - password resets should require MFA before generating a session! the user has (likely) only demonstrated access to email. ⚠️

>handlers/post_session_token.go - seems similar to password reset because a user has (likely) only demonstrated access to email. ⚠️

Sounds good, will look at putting it in these places as well.

app/data/mysql/migrations.go
```diff
@@ -69,3 +72,18 @@ func createAccountLastLoginAtField(db *sqlx.DB) error {
...
+	if mysqlError, ok := err.(*mysql.MySQLError); ok {
+		if mysqlError.Number == 1060 { // 1060 = Duplicate column name
+			err = nil
+		}
+	}
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
thenhawke: >i think it'd be fine to add the columns in a batch DDL and simplify the migrations somewhat. i'm not sure how someone could end up in a partially migrated state after once this is finished and merged, unless they ctrl+c at the magical moment during migrations.

Could I get some clarification for this? Should I combine all the migrations into one batch?

app/services/totp_creator.go
```diff
@@ -0,0 +1,48 @@
...
+		Issuer:      audience.Hostname,
+		AccountName: account.Username,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if err := totpCache.CacheTOTPSecret(account.ID, []byte(key.Secret())); err != nil {
+		return nil, errors.Wrap(err, "TOTPCreator")
+	}
```
thenhawke: > crazy thought: do we need to keep a copy on the server during this process? the best argument i can see is that it allows us to expire the attempt. it really feels like i'm missing something obvious that my past self felt more confident about. 😬

The secret is plain text and shared between the client and the server. If AuthN doesn't keep the copy on the server then technically the client could ignore the secret AuthN sends and generate their own. This would allow the client to send whatever secret they want as long as they also send a code that verifies against it. Allowing clients the ability to choose the shared secret seems a little icky to me but I'm definitely not an expert. I can't see anything in [this document](https://tools.ietf.org/html/rfc6238) that would prevent the client from generating it if the AuthN secret was ignored.

app/services/totp_creator.go
```diff
@@ -0,0 +1,48 @@
...
+
+	//Convert to png
+	var buf bytes.Buffer
+	img, err := key.Image(200, 200)
+	if err != nil {
+		return nil, err
+	}
+	png.Encode(&buf, img)
```
thenhawke: > ah, yep, that's what i'm referring to. the value being encoded with QR [happens to be a URI](https://github.com/pquerna/otp/blob/master/totp/totp.go#L198) containing information like the secret.

app/services/totp_setter.go
```diff
@@ -0,0 +1,46 @@
...
+
+	account, err := AccountGetter(accountStore, accountID)
+	if err != nil {
+		return err
+	}
+
+	secret, err := totpCache.LoadTOTPSecret(account.ID)
+	if err != nil {
+		return err
+	}
```
thenhawke: > if the secret has expired from cache, would we learn about it from this error check or the next?

app/services/credentials_verifier.go
```diff
@@ -45,5 +47,16 @@ func CredentialsVerifier(store data.AccountStore, cfg *app.Config, username stri
...
+	//Check TOTP MFA
+	if account.TOTPEnabled() {
+		secret, err := compat.Decrypt([]byte(account.TOTPSecret.String), cfg.DBEncryptionKey)
+		if err != nil {
+			return nil, errors.Wrap(err, "TOTPDecrypt")
+		}
+		if !totp.Validate(totpCode, secret) {
+			return nil, FieldErrors{{"totp", ErrFailed}}
+		}
+	}
```
thenhawke: I have this block in a few different places, should I make a function somewhere in lib or maybe a service that takes an account and an encryption key and checks if it's valid?

thenhawke:
> handlers/post_password.go - password resets should require MFA before generating a session! the user has (likely) only demonstrated access to email. ⚠️

> handlers/post_session_token.go - seems similar to password reset because a user has (likely) only demonstrated access to email. ⚠️

These are being checked now.

app/services/totp_setter.go
thenhawke: If the secret has expired I believe the read on the encrypted blob store returns an empty string. This error would catch any problems with the actual redis read which would be passed to the caller and trigger a panic in the handler. (Not a field error.) The next error would catch the empty string because the TOTP token wouldn't be valid. If the redis secret expired it would fail to validate returning a field error.

server/handlers/post_totp.go
```diff
@@ -0,0 +1,33 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		// check for valid session with live token
+		accountID := sessions.GetAccountID(r)
+		if accountID == 0 {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		if err := services.TOTPSetter(app.AccountStore, app.TOTPCache, app.Config, accountID, r.FormValue("code")); err != nil {
+			w.WriteHeader(http.StatusUnauthorized)
```
cainlevy: validation errors are typically returned with `http.StatusUnprocessableEntity` (handled by `WriteErrors`). is there a reason to prefer `http.StatusUnauthorized` here?

app/services/password_changer.go
```diff
@@ -24,5 +26,16 @@ func PasswordChanger(store data.AccountStore, r ops.ErrorReporter, cfg *app.Conf
 		return FieldErrors{{"credentials", ErrFailed}}
 	}
 
+	//Check TOTP MFA
+	if account.TOTPEnabled() {
```
cainlevy: So `POST /password` does two things. It trips me up every time.

1) `PasswordChanger` is called if the user has a session already. It's intended for use in an account settings page.
2) `PasswordResetter` is called if the user has a token instead of a session. It's intended for lost password flows.

Given that, I'm unsure about requiring TOTP here, after the user has logged in. I don't really see other sites doing it. Thoughts?

app/services/password_resetter.go
```diff
@@ -38,5 +40,16 @@ func PasswordResetter(store data.AccountStore, r ops.ErrorReporter, cfg *app.Con
 		return 0, FieldErrors{{"token", ErrInvalidOrExpired}}
 	}
 
+	//Check TOTP MFA
+	if account.TOTPEnabled() {
+		secret, err := compat.Decrypt([]byte(account.TOTPSecret.String), cfg.DBEncryptionKey)
+		if err != nil {
+			return -1, errors.Wrap(err, "TOTPDecrypt")
```
cainlevy: can you say more about `-1`? the above failures return a zero value.

app/services/password_resetter.go
```diff
@@ -38,5 +40,16 @@ func PasswordResetter(store data.AccountStore, r ops.ErrorReporter, cfg *app.Con
 		return 0, FieldErrors{{"token", ErrInvalidOrExpired}}
 	}
 
+	//Check TOTP MFA
+	if account.TOTPEnabled() {
```
cainlevy: How do you feel about requiring TOTP to successfully perform this action?

Originally I was imagining that we could perform the action but wait to establish a session until they verified TOTP, but that might require tackling the concept of a [partially authenticated session](https://github.com/keratin/authn-server/issues/92#issuecomment-481367655).

I think we can reasonably revisit this later without creating a backwards incompatibility. For example, if we figure out partially authenticated sessions then the password reset process could generate either a partial or full session depending on whether the TOTP code was provided.

app/services/password_changer.go
thenhawke: Using github as an example, if you ask for a password reset and your account has 2fa enabled, when you click on the reset link you are asked for the 2fa code before you can enter a new password. If this behavior is handled by `PasswordResetter` then the 2fa should remain on that.

If you're already logged into Github you can change your password with your current password and a new password. There's no 2fa because you've presumably already demonstrated you have the secret when you got your session logging in. If this behavior is handled by `PasswordChanger` then I think the 2fa should be removed from this endpoint. (I currently have it on there)

I think it should be removed from `PasswordChanger`, want me to do that?

server/handlers/get_totp.go
```diff
@@ -0,0 +1,35 @@
...
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		img, err := services.TOTPCreator(app.AccountStore, app.TOTPCache, accountID, route.MatchedDomain(r))
+		if err != nil {
+			panic(err)
+		}
+
+		w.Header().Set("Content-Type", "image/png")
```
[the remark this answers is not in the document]
thenhawke: Either works, whatever is preferred. Personally, I'd prefer the smuggling since I'd only have to do one call and that would reduce load times if network latency is high.

app/services/password_changer.go
cainlevy: > if you ask for a password reset and your account has 2fa enabled, when you click on the reset link you are asked for the 2fa code before you can enter a new password

okay, i'm happy to have precedent like that to follow.

> I think it should be removed from PasswordChanger, want me to do that?

👍

app/services/password_resetter.go
cainlevy: decided here: https://github.com/keratin/authn-server/pull/133#discussion_r342647440

app/services/totp_creator.go
cainlevy: 3-4kb seems acceptable (and surprising) 👍 

> Or perhaps we could default to a certain size but accept a "size" parameter from the http request to return the PNG rendered to a certain size maybe?

Yeah that'd be easy to retrofit later. Let's wait for the need before implementing custom sizes.

server/handlers/get_totp.go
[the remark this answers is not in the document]
cainlevy: hmm it's probably easier for the server to return all of the data at once rather than deal with race conditions from a pair of matched endpoints.

in javascript i think we'll be able to decode the binary PNG data and return a convenient data URI. it might be worth a proof of concept before this PR is finally merged though. i can assist with that if it's helpful.

xcskier56:
What is the status of this PR? We are using Keratin and have a client that needs some form of 2fa. Thanks!

cainlevy:
@xcskier56 I think the biggest obstacle right now is a reference client to settle questions around API design. I'd love to refresh and merge this code, but my recent sites haven't required 2fa which means I haven't had a UI prototype to confidently target.

What's your situation? Maybe one path forward would be a prerelease from this branch that you could use to give feedback on usability. What would you need to make that happen?

xcskier56:
Hey @cainlevy, I was a little mistaken and it seems like the 2fa requirement is a little further down the road than I thought. That being said, it's coming, eventually, and we are going with Keratin for this and another project so I want to keep this moving forward. Unfortunately I have very little (none) Go experience, so please excuse any dumb questions that I should have been able to figure out from the code.

### General Questions/Comments
1. How "done" is this PR? I've noticed a few comments where action was going to be done but looks like it wasn't: https://github.com/keratin/authn-server/pull/133/files#r346052239, https://github.com/keratin/authn-server/pull/133/files#r346057676. Is this at a point where we could begin testing it from this branch?
2. What is the public API implementation to require TOTP?
3. For the public API, it looks like TOTP is a required on `POST /session` and `POST /password`. How should the client know to prompt the user for the TOTP? 

### Use case thoughts/questions:
1. Is it possible to force MFA for every user? Or, alternatively, set a flag on the Account requiring MFA setup on first login? One of our use cases is a SAAS product where users would be enrolled by an admin. My understanding of this PR is that we only check the TOTP code if TOTP is enabled. 
2. What do you think about SMS 2fa? It is certainly less secure, and less ideal in many situations, but when it is required for all accounts, it is simpler to setup and enroll users in bulk.

Thanks!  -- Doug

cainlevy:
@xcskier56 I expect that requiring MFA will need to be a separate implementation. I'd be happy to discuss further in a new feature request (issue). 

SMS 2FA will also need to be separate. The biggest challenge I see is getting a sense of what providers people would expect to be able to use. I only have experience with Twilio.

@jeffreylo has rebased and kept this work alive in a follow-up PR #183. :tada: I think your general questions can be discussed further over there.

Issue #135, opened 2019-10-10
Question: Disable signup-endpoint


ppacher:
Hi,
is it possible to just disable the signup endpoint? I'd like to use authn-server but without a public signup. I know I can route everything through an application gateway but this requires to ensure that all other routes are correctly forwarded and I'd like to terminate TLS just before the authn-server. 

All users would be created by "administrators" using the private ImportAccount endpoint (or the signup endpoint if I can configure it to be "private").
Thanks in advance

Mohammed90:
Set the `ENABLE_SIGNUP` env var to false

https://github.com/keratin/authn-server/blob/master/app/config.go#L212

ppacher:
Awesome thanks! Seems this one is missing in the documentation.

PR #136, opened 2019-10-10
Added documentation for ENABLE_SIGNUP

ppacher:
Update documentation to include `ENABLE_SIGNUP` so #135 does not happen again :).

I guess a simple write-up may also be a good idea but documenting the environment variable should be enough in the first step


Mohammed90:
I think `0` is also among the falsey values. Can you check, please? I'm on my phone now, otherwise I would've checked.

cainlevy:
Thanks y'all for catching this one!

PR #145, opened 2020-01-29
Add support for application/json encoded POST/PUT/PATCH requests

### LINKED ISSUE #143: AuthN does not parse JSON requests
AuthN does not parse requests with `Content-Type: application/json`. This can be a hurdle to getting started for folks testing the server directly without the aid of a client library.

martinreus:
#### What 
Fixes #143 

#### How
Create a Parser that decides from where to read input from user, either JSON body or request form values.


martinreus:
I just realized this can be done much easier because we don't need to unmarshal to a structure.. we may simply return a map.. will do the changes

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## HEAD
 
+### Added
+
+* Support `Content-Type: application/json` [#143]
```
martinreus: I did not put it under ## 1.8.0 version because I don't know if / when a new version is going to be released :smiley:

martinreus:
Added the docs and changes in CHANGELOG; in CHANGELOG however, I didn't know exactly where to add the changes =)

CHANGELOG.md
cainlevy: HEAD is perfect 👍

PR #149, opened 2020-02-08
Add password score endpoint

martinreus:
## What
Add endpoint for checking password's zxcvbn score

## Motivation
Having to use a client JS library to check for password score adds almost 800kb payload that has to be downloaded by the user. Even though registration happens only once, having to download this amount of data on a sometimes slower mobile connection could very much frustrate users. Moving this check to the backend also ensures that the algorithm (and dictionaries used for the zxcvbn) is the same.


martinreus:
> Makes sense to me! We did [notice a difference](https://github.com/keratin/authn-server/pull/99) in zxcvbn libraries once.
> 
> Would you add an entry to `docs/api.md` and a mention in `docs/guide-displaying_a_password_strength_meter.md`?

So sorry, forgot the docs again :sweat_smile: 
Done!

PR #152, opened 2020-02-29
Add SESSION_DOMAIN configuration

martinreus:
### What
Add the possibility to define `Domain` configuration for the session cookie. 

### Why
Imagine a [SPA](https://en.wikipedia.org/wiki/Single-page_application) running under `example.com` and Authn server running under `authn.example.com`. Up until now, authenticating a user means that a session token will be set exclusively for `authn.example.com`. If this SPA has Server Side Rendering capabilities, there is no way of sending this session token to the server when accessing through `example.com`, since cookie was set for authn domain only. 
This prevents the server from rendering the page as if the user was logged in, since there is no access to said session.

### Restrictions
Setting the Domain will only work if authn is running on the same top domain as the web application, for instance `authn.example.com` and `example.com`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## HEAD
 
 * Added endpoint for checking zxcvbn password score
+* Added `SESSION_DOMAIN` configuration to allow setting session on another domain.
 
 ## 1.8.0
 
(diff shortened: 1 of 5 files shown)


martinreus:
Hmm fair point..
But isn't the app already in the security perimeter if it uses the id_token as the auth token?

cainlevy:
Currently the host app doesn't have any of the secrets necessary to create the session tokens. It can verify the auth token, yes, but only using public key cryptography that it doesn't need to store or secure. This doesn't absolve the host app of all security concerns, but it does move at least one concern somewhere else.

Happy to continue discussing, but this iteration doesn't look like the right fit for AuthN yet.

PR #155, opened 2020-05-08
Add microsoft oauth authentication

banux:
Oauth2 authentication for Microsoft, it can be useful for public facing or private domain link to office 365.


cainlevy:
Released as v1.10.0

Issue #168, opened 2021-03-02
Permission Error


borag:
I'm getting ./authn: Permission denied with docker latest version.

cainlevy:
Looks like the file is not executable. Might be an issue in the new build process.

Workaround for this affected version is to add `chmod u+x authn` to the command string. For example:

`sh -c "chmod u+x authn && ./authn server"`

cainlevy:
fixed in latest and in 1.10.3

PR #171, opened 2021-03-23
fixes for the release process

cainlevy:
- embed the release version correctly
- restore executable permissions from artifact storage before building Docker image
- remove unnecessary bzip compression step


cainlevy:
Fixes #168

PR #176, opened 2021-04-26
Update AUTHN_URL Documentation

xcskier56:
In setting up keratin, I have found that if you do not include the scheme with your `AUTHN_URL`, the application will only return 404 errors. The fact that it returns 404s for everything if this is not configured properly seems like a good task to investigate

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -19,7 +19,7 @@
 |           |    |
 | --------- | --- |
 | Required? | Yes |
-| Value | URL |
+| Value | URL. Must include scheme (http:// or https://) |
 
 This specifies the base URL of the AuthN service. It will be embedded in all issued JWTs as the `iss`. Clients will depend on this information to find and fetch the service's public key when verifying JWTs.
 


kkohrt:
@cainlevy 
When the [code in config.go](https://github.com/keratin/authn-server/blob/master/app/config.go#L111-L124) ingests the `AUTHN_URL` it uses your `lookupURL` function that not only reads in the string value of `AUTHN_URL`, but [also parses it](https://github.com/keratin/authn-server/blob/master/app/env.go#L40-L42)  in order to return a URL object; 

Unfortunately, the [URL library parse function](https://golang.org/pkg/net/url/#Parse) quietly fails you, as stated in their doc:
```
Trying to parse a hostname and path without a scheme is invalid 
but may not necessarily return an error, due to parsing ambiguities.
```
As a  result, you not only do not have a scheme defined, you do not even get the hostname (probably because of those same "parsing ambiguities.") Some additional testing has shown that adding ports or paths without a scheme can get you even worse results, as shown using this test program:
https://gist.github.com/kkohrt/65f6e8656b0a7810db7bb6394595f849

In any case, the bad parsing means that no errors are thrown, but it totally messes up your [Matches and FindDomain functions](https://github.com/keratin/authn-server/blob/b2ee0bcef67dcf3f6b3881d5680373079031c64a/lib/route/domain.go#L24-L46) which would, at least, cause your [OriginValidator](https://github.com/keratin/authn-server/blob/master/server/cors/origin_validator.go) to return false. I do not know how the validators work in the middleware, but I'd be willing to bet that "false" from a validator causes some failure mode--perhaps a 404?

So--maybe it is better to use [ParseRequestURI](https://golang.org/pkg/net/url/#ParseRequestURI) and let it throw an error when parsing config values? 

Or maybe manually patch up anything missing a scheme by prepending a http? Or use https if, for some reason, port 443 is explicitly specified? (although I could see that could cause other gotcha's down the line)

If you have a preference, we could try a more functional PR; but I'd hate to write all the code (and tests!) for the wrong solution. I'd also hate to write any code if documentation suffices :-)

cainlevy (approved):

PR #177, opened 2021-06-09
Update docs to point at new fork of zxcvbn

xcskier56:
I found that there is a new fork of the zxcvbn library for the frontend. This has some nice features like optionally not including the dictionaries to reduce file size.


cainlevy:
Nice, I sure do appreciate modernized and supported frontend libraries. 😂 

The dilemma here is that the backend's zxcvbn library specifically promises compatibility with v4.4.2 of the dropbox original. If someone installs a client library with a different scoring mechanism, the divergence can cause ([and has caused](https://github.com/keratin/authn-server/pull/99)) problems.

Based on https://github.com/dropbox/zxcvbn/issues/290#issuecomment-754791316 it looks like zxcvbn-ts prefers to support new features rather than promise compatibility. That's a reasonable choice, but it creates a new moving target for backend libraries that want to provide compatibility.

Until there is a backend golang port of zxcvbn-ts, I don't think I can officially recommend it. Cross-platform compatibility is the basis for this guide.

PR #183, opened 2021-08-20
Add TOTP multi-factor authentication support

jeffreylo:
An attempt to see https://github.com/keratin/authn-server/pull/133 through. Squashed and rebased against v1.11.0.

My understanding is that we may need a way to try this out before we feel comfortable merging — is that right, @cainlevy?

### DIFF
--- a/app/services/credentials_verifier.go
+++ b/app/services/credentials_verifier.go
@@ -4,7 +4,9 @@ import (
 	"github.com/keratin/authn-server/app"
 	"github.com/keratin/authn-server/app/data"
 	"github.com/keratin/authn-server/app/models"
+	"github.com/keratin/authn-server/lib/compat"
 	"github.com/pkg/errors"
+	"github.com/pquerna/otp/totp"
 	"golang.org/x/crypto/bcrypt"
 )
 
@@ -15,7 +17,7 @@ var emptyHashes = map[int]string{
 	12: "$2a$12$w58M3IGXURRAqXQ/OAsMmuqcV4YqP3WyJ.yHvHI5ANUK1bRWxeceK",
 }
 
-func CredentialsVerifier(store data.AccountStore, cfg *app.Config, username string, password string) (*models.Account, error) {
+func CredentialsVerifier(store data.AccountStore, cfg *app.Config, username string, password, totpCode string) (*models.Account, error) {
 	if username == "" && password == "" {
 		return nil, FieldErrors{{"credentials", ErrFailed}}
 	}
@@ -43,6 +45,17 @@ func CredentialsVerifier(store data.AccountStore, cfg *app.Config, username stri
 	}
 	if account.RequireNewPassword {
 		return nil, FieldErrors{{"credentials", ErrExpired}}
+	}
+
+	//Check TOTP MFA
+	if account.TOTPEnabled() {
+		secret, err := compat.Decrypt([]byte(account.TOTPSecret.String), cfg.DBEncryptionKey)
+		if err != nil {
+			return nil, errors.Wrap(err, "TOTPDecrypt")
+		}
+		if !totp.Validate(totpCode, secret) {
+			return nil, FieldErrors{{"totp", ErrInvalidOrExpired}}
+		}
 	}
 
 	return account, nil
--- a/app/services/password_changer.go
+++ b/app/services/password_changer.go
@@ -8,7 +8,7 @@ import (
 	"golang.org/x/crypto/bcrypt"
 )
 
-func PasswordChanger(store data.AccountStore, r ops.ErrorReporter, cfg *app.Config, id int, currentPassword string, password string, totpCode string) error {
+func PasswordChanger(store data.AccountStore, r ops.ErrorReporter, cfg *app.Config, id int, currentPassword string, password string) error {
 	account, err := store.Find(id)
 	if err != nil {
 		return errors.Wrap(err, "Find")
--- a/app/services/password_changer.go
+++ b/app/services/password_changer.go
@@ -3,12 +3,14 @@ package services
 import (
 	"github.com/keratin/authn-server/app"
 	"github.com/keratin/authn-server/app/data"
+	"github.com/keratin/authn-server/lib/compat"
 	"github.com/keratin/authn-server/ops"
 	"github.com/pkg/errors"
+	"github.com/pquerna/otp/totp"
 	"golang.org/x/crypto/bcrypt"
 )
 
-func PasswordChanger(store data.AccountStore, r ops.ErrorReporter, cfg *app.Config, id int, currentPassword string, password string) error {
+func PasswordChanger(store data.AccountStore, r ops.ErrorReporter, cfg *app.Config, id int, currentPassword string, password string, totpCode string) error {
 	account, err := store.Find(id)
 	if err != nil {
 		return errors.Wrap(err, "Find")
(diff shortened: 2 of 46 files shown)


cainlevy:
Yes that's right. I will make time to review and refresh my memory on what the next steps are. Maybe the best way to handle this is with a pre-release, or just advertise it as unstable until it's been implemented in a production environment at least once. 🤔

server/handlers/get_totp.go
```diff
@@ -0,0 +1,35 @@
...
+func GetTOTP(app *app.App) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// check for valid session with live token
+		accountID := sessions.GetAccountID(r)
+		if accountID == 0 {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		img, err := services.TOTPCreator(app.AccountStore, app.TOTPCache, accountID, route.MatchedDomain(r))
```
cainlevy, line 24: This has been a question mark for me [since the original PR](https://github.com/keratin/authn-server/pull/133#discussion_r331208949), because it means that a GET request is generating server state. It also involves create a new TOTPCache, adding `.write()` semantics to the BlobStore, etc., and I wonder if it could be simpler.

What if GetTOTP generated a new secret every time, and returned it in a JWT alongside the QR code? The PostTOTP endpoint would validate the JWT, extract its secret, and use that to verify the user has a working code generator.

Thoughts? It feels consistent with how AuthN handles other processes. In fact, it's similar enough to password resets that I wonder if we would need to imitate the locking mechanism there and ensure that no matter how many JWTs are generated, using one will expire the rest.

server/handlers/get_totp.go
```diff
@@ -0,0 +1,35 @@
...
+		if err != nil {
+			panic(err)
+		}
+
+		w.Header().Set("Content-Type", "image/png")
+		w.Header().Set("Content-Length", strconv.Itoa(len(img.Bytes())))
```
cainlevy, line 30: I would like to revisit this response so that it can be JSON containing a PNG. In the [previous discussion](https://github.com/keratin/authn-server/pull/133#discussion_r334215025) I imagined returning the URI as well for implementations that don't want to depend entirely on QR code scanning. Whether or not we return the URI or refactor this endpoint to create a JWT, a JSON response would give us more flexibility to revisit response requirements at a later date.

app/services/totp_deleter.go
```diff
@@ -0,0 +1,20 @@
...
+//TOTPDeleter removes TOTP from the specified account
+func TOTPDeleter(accountStore data.AccountStore, accountID int) error {
+	//Delete totp secret in database
+	affected, err := accountStore.DeleteTOTPSecret(accountID)
+	if err != nil {
+		return errors.Wrap(err, "TOTPDeleter")
+	}
+	if !affected {
+		return errors.New("unable to delete totp secret")
+	}
```
cainlevy, line 17: should we handle this gracefully? i can't think of downsides to returning nil in this case.

app/services/credentials_verifier.go
```diff
@@ -45,5 +47,16 @@ func CredentialsVerifier(store data.AccountStore, cfg *app.Config, username stri
...
 	}
 
+	//Check TOTP MFA
+	if account.TOTPEnabled() {
+		secret, err := compat.Decrypt([]byte(account.TOTPSecret.String), cfg.DBEncryptionKey)
+		if err != nil {
+			return nil, errors.Wrap(err, "TOTPDecrypt")
+		}
+		if !totp.Validate(totpCode, secret) {
+			return nil, FieldErrors{{"totp", ErrInvalidOrExpired}}
```
cainlevy, line 57: @xcskier56 you asked how a client could know that someone needs to supply a token. i expect that it will be a guess-and-test process, with this error signaling the extra requirement.

1. collect and submit username/password
2. catch totp error
3. reveal otp input and merge with username/password in a second submit

app/services/password_changer.go
```diff
@@ -24,5 +26,16 @@ func PasswordChanger(store data.AccountStore, r ops.ErrorReporter, cfg *app.Conf
...
+	if account.TOTPEnabled() {
+		secret, err := compat.Decrypt([]byte(account.TOTPSecret.String), cfg.DBEncryptionKey)
+		if err != nil {
+			return errors.Wrap(err, "TOTPDecrypt")
+		}
+		if !totp.Validate(totpCode, secret) {
+			return FieldErrors{{"totp", ErrInvalidOrExpired}}
+		}
+	}
+
```
cainlevy, line 39: we decided in the previous PR to remove this requirement, since PasswordChanger is called when users already have an authenticated session.

https://github.com/keratin/authn-server/pull/133#discussion_r342647440

server/public_routes.go
```diff
@@ -34,6 +34,18 @@ func PublicRoutes(app *app.App) []*route.HandledRoute {
 		route.Get("/session/refresh").
 			SecuredWith(originSecurity).
 			Handle(handlers.GetSessionRefresh(app)),
+
+		route.Get("/totp").
```
cainlevy, line 38: thoughts on `GET /totp/new`?

as it is currently, i might expect to get information about totp status.

cainlevy (changes requested):
okay! thank you for picking this up and championing it. i've regained context and i'm excited to see it through.

we'll want to make sure we have documentation before releasing, on both an API level (new error codes, new params, new endpoints) and a "how to" level. that documentation can have disclaimers about the functionality being subject to change until a client has successfully implemented the new protocol.

app/services/credentials_verifier.go
alee792: This may have come up in previous discussion, but if other authentication factors are on the roadmap, it may make sense to have an explicit API to signal required factors. 

I also worry that coupling TOTP and basic credentials into the same call makes OIDC ACR-like "step-up" authentication more difficult (certain endpoints only require credentialed authn, but others may require elevated factors). That may be out of scope for Keratin; i.e. Keratin is one among many identity providers that can satisfy factors, not the application that assembles and verifies them.

cainlevy: > it may make sense to have an explicit API to signal required factors

Makes sense to me. That seems like an easy addition. Do you think it would require changes at this step?

> "step-up" authentication more difficult (certain endpoints only require credentialed authn, but others may require elevated factors)

I've been assuming Keratin would need to handle this for installations where 2FA becomes mandated. I hadn't considered exposing the half-authenticated state to the host application, but I think it'd be an easy addition with a new JWT claim. Thoughts?

docker-compose.yml
```diff
@@ -6,6 +6,7 @@ services:
       - "${REDIS_PORT:-8701}:6379"
 
   mysql:
+    platform: linux/x86_64
```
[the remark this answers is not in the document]
cainlevy: fixed in main

app/services/totp_setter.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"github.com/keratin/authn-server/app"
+	"github.com/keratin/authn-server/app/data"
+	"github.com/keratin/authn-server/lib/compat"
+	"github.com/pkg/errors"
+	"github.com/pquerna/otp/totp"
+)
+
+//TOTPSetter persists the TOTP secret to the accountID if code is correct
```
cainlevy, line 11: what is the correct behavior if a totp secret is already persisted?

server/handlers/get_totp.go
```diff
@@ -0,0 +1,46 @@
...
+func GetTOTP(app *app.App) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// check for valid session with live token
+		accountID := sessions.GetAccountID(r)
+		if accountID == 0 {
+			w.WriteHeader(http.StatusUnauthorized)
+			return
+		}
+
+		totpKey, err := services.TOTPCreator(app.AccountStore, app.TOTPCache, accountID, route.MatchedDomain(r))
```
cainlevy, line 25: resurfacing: https://github.com/keratin/authn-server/pull/183#discussion_r697918014

PR #187, opened 2021-11-17
Update docs to match zxcvbn password policy scale

Harwood:
Correct docs to show possible values for `PASSWORD_POLICY_SCORE` to be `0 - 4` instead of `0 - 5` to be in line with range used by [zxcvbn](https://github.com/dropbox/zxcvbn#usage).


cainlevy:
Egads, yes!

Issue #156, opened 2020-05-20
More flexible domain restrictions


zekenie:
It would be wonderful if we could use wildcards in the `APP_DOMAINS` environment variable. We are trying to enable staging deploys for each pull request to our frontend. These frontends all point to the same backend staging server and authn instance. I'd love `pr-*.ourdomain.com` to be allowed to make requests to authn.

cainlevy:
This looks quite achievable. I think nearly all of the functionality could be built into one function:

https://github.com/keratin/authn-server/blob/e1b22d8acc39b2ec618136792f53c76c1c7790c2/lib/route/domain.go#L39-L63

That should enable CORS and the Origin validation to work. The remaining bit that I've found is a failsafe return URL for OAuth:

https://github.com/keratin/authn-server/blob/e1b22d8acc39b2ec618136792f53c76c1c7790c2/server/handlers/get_oauth.go#L19-L26

This code path isn't expected though, unless something is misconfigured or a user is messing around with URLs. A simple word of advice in the documentation to maintain a non-wildcard domain in first position should be enough.

@zekenie Is this something you're interested in implementing?

zekenie:
@cainlevy I'd love to. Unfortunately I've never touched go before, so I'm a bit concerned I'll get stuck on simple stuff. I might give it a try this weekend. If I can't figure it out, I'll just enthusiastically cheer from the sidelines in the hopes that you or someone else picks it up

PR #189, opened 2021-12-25
Add support for flexible app domains

jeffreylo:
This patch adds wildcard support in APP_DOMAINS, which enables users to configure authn-server to be more flexible when validating authorised hosts.

Closes https://github.com/keratin/authn-server/issues/156.


lib/route/domain.go
```diff
@@ -80,3 +80,34 @@ func (d *Domain) URL() url.URL {
 	}
 	return url.URL{Scheme: "http", Host: d.String()}
 }
+
+// Match returns true if the text satisfies the pattern
+// string, supporting only '*' wildcard in the pattern.
+func Match(pattern, name string) bool {
+	if pattern == "" {
+		return name == pattern
+	}
```
cainlevy, line 89: 🤔 i'm trying to think of a need for this. it may be a backwards compatible ... bug? i'd be happy returning false for empty patterns.

cainlevy (approved):
looks good!

let me know what you think of the empty case. i'll aim to merge in a few days if nothing changes.

lib/route/domain.go
jeffreylo: That’s a fair point. Updated. I also took the time to un-export this function.

cainlevy (approved):

cainlevy:
released in v1.12.0

Issue #186, opened 2021-10-20
Non-default Redis user does not work due to outdated version of go-redis library 


Probotect0r:
I am trying to run an AuthN Server that is backed by an AWS Elasticache Redis instance. I have created two different  Elasticache Users, which map directly to Redis ACL users. One of the users is `default`, and the other is `user1`.

When I set `REDIS_URL` to `redis://user1:password@myhost:myport`, I am not able to authenticate. When I use `redis://default:password@myhost:myport` or `redis://:password@myhost:myport`, it works correctly. 

Digging into the underlying `go-redis` library (version 6.15.2 as used by this module), It seems like the username is ignored entirely, as seen here: https://github.com/go-redis/redis/blob/v6.15.2/options.go#L174-L179. But, starting with V8 of the library, Username is not ignored: https://github.com/go-redis/redis/blob/v8.0.0/options.go#L201-L206.

Are there any plans on upgrading the `go-redis` library to 8.0 or above? It would be nice to be able to use a non-default Redis user. 

Thanks

cainlevy:
I agree, that's a good reason to update the dependency.

PR #191, opened 2021-12-30
dep: bump go-redis (v6.15.2 => v8.11.4)

jeffreylo:
This patch bumps the [go-redis](https://github.com/go-redis/redis) library to v8, which resolves an issue when authn-server is configured with a non-default Redis user.

v8 notably introduces the use of the stdlib's context package to its client functions. This patch sidesteps a decision on whether the blob and refresh token stores should also be context-aware by using `context.TODO()` as needed.

Closes https://github.com/keratin/authn-server/issues/186.


cainlevy (approved):

cainlevy:
v1.13.0

PR #198, opened 2022-04-07
Added public routes for /jwks and /configuration

Probotect0r:
Closes #197 

I haven't been able to get the dev environment working, so haven't been able to test this. Although, the changes are simple copy and paste.

Here's what I am running into (not a Go developer so please advise on anything obvious):
```
go install
# github.com/mattn/go-sqlite3
sqlite3-binding.c: In function ‘sqlite3SelectNew’:
sqlite3-binding.c:117843:10: warning: function may return address of local variable [-Wreturn-local-addr]
117843 |   return pNew;
       |          ^~~~
sqlite3-binding.c:117802:10: note: declared here
117802 |   Select standin;
       |          ^~~~~~~
ego server/views
make: ego: No such file or directory
make: *** [Makefile:14: init] Error 127
```
Seems like I also need `ego` installed, but I have run `go get github.com/go-ego/ego` without success. 

Also, the main readme mentions `make docker`, but that target doesn't exist in the Makefile.

PR #199, opened 2022-04-09
dep: record ego as tooling dependency

jeffreylo:
Reported at
https://github.com/keratin/authn-server/pull/198#issuecomment-1092440793:

> Took me a moment to uncover the problem but this appears to be a
> different "ego" library. AuthN relies on go get
> github.com/benbjohnson/ego instead. That used to be part of the
> go.mod list until a recent PR.

Add a tools.go to record the tooling dependency per guidance:
https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module. This
ensures ego sticks around when authn-server's dependencies are
modified.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.db
 .env
-vendor/
-dist/
 .idea
+bin
+dist/
+vendor/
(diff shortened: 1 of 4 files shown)


cainlevy (approved):
That's a neat solution! Thanks

PR #200, opened 2022-06-02
services/accounts: do not permit username as password

jeffreylo:
Prevent users from re-using their usernames as passwords as reasonably
complex usernames can pass zxcvbn's password scoring.


server/test/asserts.go
```diff
@@ -7,17 +7,18 @@ import (
...
 	"github.com/keratin/authn-server/app/services"
 	"github.com/keratin/authn-server/app/tokens/identities"
 	"github.com/keratin/authn-server/app/tokens/sessions"
+	"github.com/keratin/authn-server/server/handlers"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
 func AssertData(t *testing.T, res *http.Response, expected interface{}) {
+	t.Helper()
```
cainlevy, line 21: useful!

cainlevy (commented):
Could `PasswordValidator()` be a good centralized place for this logic? What were your considerations?

jeffreylo:
> Could `PasswordValidator()` be a good centralized place for this logic? What were your considerations?

Ah, yes. I originally did this change via the password validator, updating the func to take the username of the user. Realised that the one place where we _don’t_ have the username immediately is at password reset time. Waffled a bit on this. Think it’s the right decision though since [we now incur another query for the username lookup anyways](https://github.com/keratin/authn-server/pull/200/files#diff-f1a8ba37da4e29cdc083c7685d494479121ab83ba9e82c9878c6ae619e7fdfb2R25). Thanks for prompting me to reconsider!

jeffreylo:
@cainlevy — pushed the change to have the `PasswordValidator` own the username/password equality check. Thanks again for the initial review.

jeffreylo:
> thanks! i'll add a note to the CHANGELOG. do you need this in a release?

I don’t have a strong preference. Cutting a release would be nice as it can prevent an unintentional class of user error. Thanks for updating the CHANGELOG! Wasn’t sure if this was expected of contributors as there’s no explicit mention of it in [the CONTRIBUTING guide](https://github.com/keratin/authn-server/blob/main/CONTRIBUTING.md). I’ll make sure to suggest release urgency and update the changelog moving forward if that’s preferable to you 😄

PR #201, opened 2022-06-06
feat: :sparkles: Epic SMART on FHIR

lucasthomazoni:


xcskier56:
@cainlevy FYI we are working on implementing a healthcare specific extension to OIDC called SMART on FHIR. This is still a draft and probably won't ever make sense to upstream into the core Keratin server since it is so specifically healthcare focused, but wanted to let you know what we're doing with the project!

cainlevy:
thanks i will take a look!

cainlevy:
This much looks like a fairly standard OAuth integration and I would have no issues including it in AuthN. 🌟

PR #205, opened 2023-04-18
claims: add SessionID to known claims

AlexCuse:
marshaled to 'sid' JWT claim https://www.iana.org/assignments/jwt/jwt.xhtml#claims

### DIFF
--- a/app/services/session_refresher_test.go
+++ b/app/services/session_refresher_test.go
@@ -8,12 +8,14 @@ import (
 	"github.com/keratin/authn-server/app/data/mock"
 	"github.com/keratin/authn-server/app/data/private"
 	"github.com/keratin/authn-server/app/services"
+	"github.com/keratin/authn-server/app/tokens/identities"
 	"github.com/keratin/authn-server/app/tokens/sessions"
 	"github.com/keratin/authn-server/lib/route"
 	"github.com/keratin/authn-server/ops"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/square/go-jose.v2/jwt"
 )
 
 func TestSessionRefresher(t *testing.T) {
@@ -30,6 +32,7 @@ func TestSessionRefresher(t *testing.T) {
 	audience := &route.Domain{"authn.example.com", "8080"}
 	session, err := sessions.New(refreshStore, cfg, accountID, audience.String())
 	require.NoError(t, err)
+	assert.NotEmpty(t, session.SessionID)
 
 	t.Run("tracks actives while generating token", func(t *testing.T) {
 		activesStore := mock.NewActives()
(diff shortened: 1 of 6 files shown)


app/tokens/identities/identity.go
```diff
@@ -43,6 +43,7 @@ func New(cfg *app.Config, session *sessions.Claims, accountID int, audience stri
 			Audience: jwt.Audience{audience},
 			Expiry:   jwt.NewNumericDate(time.Now().Add(cfg.AccessTokenTTL)),
 			IssuedAt: jwt.NewNumericDate(time.Now()),
+			ID:       session.ID,
```
cainlevy, line 46: I looked through https://www.iana.org/assignments/jwt/jwt.xhtml#claims to see if any well-known claims would be a good fit. What do you think of `sid`?

app/tokens/sessions/session.go
```diff
@@ -70,6 +71,7 @@ func New(store data.RefreshTokenStore, cfg *app.Config, accountID int, authorize
 			Subject:  string(refreshToken),
 			Audience: jwt.Audience{cfg.AuthNURL.String()},
 			IssuedAt: jwt.NewNumericDate(time.Now()),
+			ID:       uuid.NewString(),
```
cainlevy, line 74: UUID is also my preference. 👍

cainlevy (commented):
Looks like a good direction to me!

app/tokens/identities/identity.go
AlexCuse: I like it kind of surprised ID in the `jwt` package doesn't end up there.  Followed same pattern as `AuthTime` in this commit: https://github.com/keratin/authn-server/pull/205/commits/219ca73a645dc45828c1aec024d0e33d4a0280ac 

I will open PR in the go client with same - and put a comment about the embed with a link to the file here 😄

AlexCuse:
moving original description here to make room for something merge worthy: 


> We are building MFA into our system that runs alongside authn-server and looking to deal with session refreshes.  I was planning to submit a PR with an optional webhook notification that includes the old and new token values but in the process I noticed that `ID` in the token claims is unused - I think if we assign a value on session creation and then thread it through on refresh, we could achieve the same thing on our end without the additional call (just cache MFA status per ID instead of per token).
> 
> Open to suggestion on other ways to assign a unique ID instead of bringing in the UUID package, that felt heavy handed but wanted to get discussion on the approach started quickly.  If this seems viable I will look to harden in general, if I am missing something I can go back to the webhook notification path.

cainlevy (approved):
Thanks!

AlexCuse:
I added a test for session creation / parsing at least.  Let me know if you have any tips on setup for `TestSessionRefresher` - just need to setup properly to parse the generated token and validate it carries through.

Also squashed down to a single commit to prep for merge.

app/services/session_refresher_test.go
```diff
@@ -41,6 +42,8 @@ func TestSessionRefresher(t *testing.T) {
 		assert.NoError(t, err)
 		assert.NotEmpty(t, identityToken)
 
+		// TODO: parse token and validate SessionID still set - what setup is needed?
```
cainlevy, line 45: have a look in server/test/asserts.go at AssertIDTokenResponse. i doubt it will be useful in whole, but it should provide an example of parsing and asserting on the JWT.

AlexCuse: Ah that's nice I was trying to think of how to use the key setup for key store, didn't think to go at it through JWT library directly.  Will give a go hopefully one more force push.  

Need to spend a little time tomorrow to build a docker image locally but the change itself feels very straightforward don't anticipate any problems as long as we deploy the client change with our MFA stuff.

AlexCuse: Yup that worked perfectly thanks - just force pushed the last test update.

cainlevy (approved):

cainlevy:
Released in v1.16.0

Issue #202, opened 2022-09-27
Access token fails to refresh in Progressive Web App (PWA) on mobile (Android and iOS), with refresh token unexpired


drevets:
Hi -- 

We've been using Authn (both authn-server and authn-js) for several years. We've discovered an issue (which is frustratingly vague), whereby users on our PWA on mobile find they have to log in _too much_. We had not changed the configuration for either the ACCESS_TOKEN_TTL or the REFRESH_TOKEN_TTL, so my understanding is that users should remain logged in for 30 days, with authn-js automatically attempting to refresh their access token at the half life (which would be every 30 min). 

The refresh attempts appear to be failing with a 401, somewhat reliably, in the PWA after about a day or a period longer than 1 hr -- I'm not exactly certain how long it takes for it to happen as I haven't been able to reliably reproduce the issue. I don't know for sure, but I don't think it matters if the user closes out of the app, or merely minimizes it and uses other applications on their phone. However, I do think that if our PWA were open on their phone, the refresh logic would work, but something about minimizing it or having it go "dormant" seems to cause an issue. 

There was a bit of documentation for the /session/refresh route that potentially seemed relevant: 

```As long as a device remains logged in to the AuthN server, it can hit this endpoint to fetch a fresh JWT session. The [keratin/authn-js](https://github.com/keratin/authn-js) library can automate this by preemptively refreshing tokens when they reach halflife.``` 

Is the mobile device not remaining logged in to the AuthN server and that's causing the problem? If so, is there a way to prevent its disconnection? 

I'm also wondering if this could be an issue related to switching networks or losing connectivity in general, though it seems like AuthN should still be able to refresh the token after losing connectivity, so long as the refresh token were still valid. 

It seems improbable that the refresh tokens themselves are expiring, given they are supposed to be good for 30 days -- so I'm also wondering if there are other reasons for the refresh attempt to fail with a 401 that might have to do with timeouts, etc. 

Versions of AuthN: 

- authn-js: 1.3.0 
- authn-server: 1.10.2

Example of failing request (from logs in Heroku): 
- "GET /session/refresh HTTP/1.1" 401 0 "https://frank-staging.netlify.app/home" "Mozilla/5.0 (iPhone; CPU iPhone OS 15_6_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.6.1 Mobile/15E148 Safari/604.1"

cainlevy:
[...]

PR #208, opened 2023-05-07
session: add option to use explicit expiry for refresh token cookies

AlexCuse:
Possibly related to #202


AlexCuse:
Need to update docs: https://github.com/keratin/authn-server/blob/main/docs/config.md

AlexCuse:
Went with `REFRESH_TOKEN_EXPIRY` to try and align with what's there (preview is from ide, other booleans render same way 😄 ).  This should be good to go I've built it locally and tested.

image

PR #211, opened 2023-06-28
feat: add HMAC signatures to webhook notifications (fixes #207)

### LINKED ISSUE #207: Notification Signing
I’d like to have a signing facility for the notifications sent on password resets / reset notifications.  A simple HMAC of the body would do for our needs but wanted to ask here before looking too close at a PR if there’s anything else we should look to support.

Not something I’ve used but I see a url configured for passwordless logins wonder if it could maybe leverage same.

AlexCuse:


app/config.go
```diff
@@ -637,6 +639,18 @@ var configurers = []configurer{
 		}
 		return nil
 	},
+
+	// WEBHOOK_SIGNING_KEY is a hex encoded key used to sign webhook notifications using sha256-HMAC
```
AlexCuse, line 643: This should maybe be `APP_SIGNING_KEY`

app/services/webhook_sender.go
```diff
@@ -33,7 +37,7 @@ func retry(schedule []time.Duration, fn func() error) error {
 	return err
 }
 
-func WebhookSender(destination *url.URL, values *url.Values, schedule []time.Duration) error {
+func WebhookSender(destination *url.URL, values *url.Values, schedule []time.Duration, key []byte) error {
```
AlexCuse, line 40: is it better to pass in a `hash.Hash` here to better support extension?  EG a function on config builds it based on the key (and eventually alg etc) specified in configuration?

app/services/webhook_sender.go
```diff
@@ -42,8 +46,17 @@ func WebhookSender(destination *url.URL, values *url.Values, schedule []time.Dur
 		Timeout: 10 * time.Second,
 	}
 
-	err := retry(schedule, func() error {
-		res, err := c.PostForm(destination.String(), *values)
+	req, err := http.NewRequest(http.MethodPost, destination.String(), strings.NewReader(values.Encode()))
```
cainlevy, line 49: what do you think ought to be done to make sure that this signing process can be easily repeated on the other side? do we need to alphabetize the keys or anything? i'm thinking about languages that may not guarantee hash order, for example.

AlexCuse: I wondered about this as well, more used to sending json and more structured formats.  It turns out the Encode() function does sort by key (https://pkg.go.dev/net/url#Values.Encode) which I figured is good enough but you're right it's not obvious.  We should document beyond unit tests somehow.

AlexCuse: A walkthrough like this would probably be appropriate.  I'll find it and put something similar together this week. 

https://keratin.github.io/authn-server/#/guide-implementing_forgotten_passwords

AlexCuse: added an example here @cainlevy [bcb2980](https://github.com/keratin/authn-server/pull/211/commits/bcb2980a9922e497bf67ca31ce7a7cd22f9222e8)

cainlevy: I think it gets the job done. It could be more idiomatic but I think eventually the AuthN clients can provide simple verification functions and the guide will be mostly directional.

cainlevy (approved):

app/services/webhook_sender.go
cainlevy: https://github.com/keratin/authn-demo might be a good start

AlexCuse: Need more time to get local environment setup right for ruby but I was able to verify the sorting approach generates the correct signature using the following in replit: 

```
require "openssl"

# add token first to force "wrong" default in order-guaranteed hashes
params = { "token" => "my token", "account_id" => "999999" }

query_string = params.map { |key, value| "#{key}=#{value}" }.join('&')
sorted_query_string = params.sort.to_h.map { |key, value| "#{key}=#{value}" }.join('&')

hex_secret_key = "my key"
secret_key = [hex_secret_key].pack("H*")

digest = OpenSSL::Digest.new("sha256")
hmac = OpenSSL::HMAC.hexdigest(digest, secret_key, query_string)
sorted_hmac = OpenSSL::HMAC.hexdigest(digest, secret_key, sorted_query_string)


puts "unsorted"
puts query_string
puts hmac

puts "\sorted"
puts sorted_query_string
puts sorted_hmac
```

Will plan to merge later today.  Maybe will take one pass at simplifying the example.

PR #213, opened 2023-07-02
tooling: add errcheck linter

AlexCuse:
I think this may be the most useful linter for catching bugs.  Tried to do the reasonable thing on violations.

### DIFF
--- a/app/services/identity_reconciler.go
+++ b/app/services/identity_reconciler.go
@@ -2,11 +2,12 @@ package services
 
 import (
 	"encoding/hex"
+
 	"github.com/keratin/authn-server/app"
 	"github.com/keratin/authn-server/app/data"
+	"github.com/keratin/authn-server/app/models"
 	"github.com/keratin/authn-server/lib"
 	"github.com/keratin/authn-server/lib/oauth"
-	"github.com/keratin/authn-server/app/models"
 	"github.com/pkg/errors"
 	"golang.org/x/oauth2"
 )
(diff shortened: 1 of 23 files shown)


lib/compat/marshal.go
```diff
@@ -69,6 +69,8 @@ func UnmarshalInt(data []byte) (int, error) {
 	}
 
 	var num int32
+	// ignored to match ruby behavior
```
AlexCuse, line 72: Should confirm this is the case.

app/services/identity_reconciler.go
```diff
@@ -62,6 +63,11 @@ func IdentityReconciler(accountStore data.AccountStore, cfg *app.Config, provide
 	if err != nil {
 		return nil, errors.Wrap(err, "AccountCreator")
 	}
-	accountStore.AddOauthAccount(newAccount.ID, providerName, providerUser.ID, providerToken.AccessToken)
+	err = accountStore.AddOauthAccount(newAccount.ID, providerName, providerUser.ID, providerToken.AccessToken)
+	if err != nil {
```
AlexCuse, line 67: This is probably the most interesting thing turned up here.  I could see an argument to continue ignoring as well.

lib/compat/marshal.go
cainlevy: The background on this ruby compatibility layer is that AuthN Server was originally built in Ruby and I wanted the Go version to be an easy upgrade. However at this point I think it's fair to say that the backwards compatibility challenge is with previous AuthN 1.x servers. Moving past this would probably require supporting two formats at once and then removing the old one in a 2.x release.

app/services/identity_reconciler.go
cainlevy: Passing it back to the caller makes sense to me. I don't see how it will happen but if it does this will be necessary to discover and fix.

cainlevy (approved):
Nice improvement!

lib/compat/marshal.go
AlexCuse: I can try to update comment before merging.  Is there a known case where this would error and we want to proceed with zero?

cainlevy: I'm not aware of any, no.

I just tried to search for how this gets used and came up short. Perhaps `UnmarshalInt` is irrelevant?

AlexCuse: I just fixed the comment for now maybe will take a pass over that file and remove anything that seems unused.  This is pretty far in the dark recesses, I certainly never saw it before and should familiarize 😄

PR #214, opened 2023-07-05
tooling: add staticcheck, unused, ineffassign linters

AlexCuse:


app/config.go
```diff
@@ -526,7 +526,18 @@ var configurers = []configurer{
 	// PORT is the local port the AuthN server listens to. The default is taken from AUTHN_URL, but
 	// may be different for port mapping scenarios as with containers and load balancers.
 	func(c *Config) error {
-		defaultPort, err := strconv.Atoi(c.AuthNURL.Port())
+		var defaultPort int
+
+		if p := c.AuthNURL.Port(); p != "" {
+			defaultPort, _ = strconv.Atoi(p)
+		} else {
+			switch c.AuthNURL.Scheme {
```
AlexCuse, line 534: It seems like this would be needed for scenarios where authn URL is configured with a default port - far as I can tell you would need to explicitly configure port for those to work which is weird.

lib/oauth/microsoft.go
```diff
@@ -25,8 +25,8 @@ func NewMicrosoftProvider(credentials *Credentials) *Provider {
 		config: config,
 		UserInfo: func(t *oauth2.Token) (*UserInfo, error) {
 			var me struct {
-				id                string
-				userPrincipalName string
+				Id                string `json:"id"`
```
AlexCuse, line 28: As far as I can tell this never would have unmarshaled.

AlexCuse:
I think this should be good to go.  `gosimple` might have had a lot of violations so not sure if I'll get as far as being able to run defaults + gosec but will probably open a PR to add to CI soon.

cainlevy (approved):
Would you make a note in the changelog for the bugs that you fixed in app/config.go and lib/oauth/microsoft.go?

AlexCuse:
Yep will do may be worth a release soonish.

PR #216, opened 2023-07-12
Update Release to build arm64 image

xcskier56:
This was a quick PR that I'm very positive will NOT work but should be a starting point for discussion. @cainlevy please let me know of any updates I should make.

Thanks!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,6 +32,32 @@ jobs:
         with:
           name: authn-linux64
           path: dist/authn-linux64
+  linux64Arm:
+    name: Compile for Linux ARM
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with:
+          go-version: ^1.17
+      - run: go mod download
+
+      - name: compile
+        run: |
+          go build \
+            -ldflags "-extldflags -static -X main.VERSION=${VERSION##*/v}" \
+            -o dist/authn-linux64arm
+        env:
+          VERSION: ${{ github.ref }}
+          GOOS: linux
+          GOARCH: arm64
+          CGO_ENABLED: 1
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: authn-linux64arm
+          path: dist/authn-linux64arm
 
   windows64:
     name: Compile for Windows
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -167,3 +203,4 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
           tags: latest,${{ steps.version.outputs.number }}
+          platform: linux/amd64,linux/arm64
```
cainlevy, line 206: It looks like the Dockerfile [only adds the authn-linux64 binary](https://github.com/keratin/authn-server/blob/main/Dockerfile#L6). You may need to make both binaries available in the build context and then somehow use [`ARG TARGETPLATFORM`](https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images) to add the correct one?

.github/workflows/release.yml
```diff
@@ -32,6 +32,32 @@ jobs:
...
+        env:
+          VERSION: ${{ github.ref }}
+          GOOS: linux
+          GOARCH: arm64
+          CGO_ENABLED: 1
+
+      - uses: actions/upload-artifact@v2
+        with:
+          name: authn-linux64arm
+          path: dist/authn-linux64arm
```
cainlevy, line 60: you might be able to verify the `linux64Arm` step by running it in your fork, grabbing the artifact, and making sure it executes as expected

xcskier56:
@cainlevy, I'm going to close this PR due to the complexity of automating an arm64 build on Github's amd based runners. Basically what I've found is that I'll have to use Qemu + docker buildx to do the actual build. On my intel mac, this takes nearly 10 minutes. Then to fit into the current build pipeline without major refactoring, you would have to extract the built authn binary from the image and upload as an artifact. 

One route that I found that would help built time is using buildJet to get an ARM runner, but this is a paid service. 

To work around this, I'm going to build our keratin images by hand and store in our own ECR, and when we need to update I'll update the image. 

If/when Github gets their act together and adds ARM runners, I might reopen and finish the implementation.

Thanks,

cainlevy:
Understood! GitHub really does need better runners.

AlexCuse:
@xcskier56 have you looked much at what the cost would be for the BuildJet runners?  I have enough interest in ARM image availability I might be willing to sponsor, especially if it's pure pay-per-minute since we don't release that often.  Self hosting ECR is enough of a pain I think offering an official ARM image could be a worthwhile addition.

AlexCuse:
If it's that cheap I may look into getting signed up over next few weeks.  I could probably use their runners for some other projects too.  There aren't a ton of weaknesses for GitHub CI but this one makes me pull my hair out.

xcskier56:
It is really cheap and took me like 5 minutes to get it running. 

1. Signup with github
2. Setup billing (or use their $5 free)
3. Change the runner to one of their runners

The place you'll likely run into issues is making sure Go is installed on the machine. We use Ruby installed directly on the runner which took me a bit to get going. However, if you use a docker image for building then this might just work. 

One other thing I ran into is that sometimes Docker hub rate-limits pulls for buildjet. All you have to do is add some credentials so that you're not pulling images anonymously. This is in their documentation

PR #217, opened 2023-07-18
tooling: add gosimple + govet linters

AlexCuse:
Pretty simple stuff in this one.  The unkeyed fields one is something I've seen burn on a few struct changes.

This gets us to the point of running all default linters + gosec which I think is a good place to be in general.  I've never found a great reason not to keep the defaults enabled even if you need to add a few `// nolint` comments once in awhile.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,9 +1,3 @@
 linters:
-  disable-all: true
   enable:
     - gosec
-    - errcheck
-    - staticcheck
-    - unused
-    - ineffassign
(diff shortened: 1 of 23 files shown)


cainlevy (approved):

.golangci.yml
```diff
@@ -1,8 +1,3 @@
 linters:
-  disable-all: true
```
cainlevy, line 2: this is a nice milestone!

PR #221, opened 2023-11-05
deps: switch to modernc.org/sqlite3 for pure go builds

AlexCuse:
For evaluation with #129 

Another pure go sqlite client that may be worth considering is https://github.com/glebarez/go-sqlite


AlexCuse:
Note the actual builds are not yet changed to remove CGO dependency.  Validated using the docker local build and an additional make target for tests since -race requires running with CGO.

go.mod
```diff
@@ -18,7 +18,7 @@ require (
 	github.com/jmoiron/sqlx v0.0.0-20170430194603-d9bd385d68c0
 	github.com/joho/godotenv v1.2.0
 	github.com/lib/pq v1.10.0
-	github.com/mattn/go-sqlite3 v1.6.0
+	github.com/mattn/go-sqlite3 v1.14.18 // indirect
```
AlexCuse, line 21: Imported for tests per https://pkg.go.dev/modernc.org/sqlite#section-readme

AlexCuse (commented):
This was interesting to look at.  Not sure how much easier it makes ARM builds if at all.  But I was impressed with the SQLite implementation in general.

PR #236, opened 2024-01-07
oauth: get signing key from provider

AlexCuse:


cainlevy (approved):

cainlevy (approved):
Looks fine to me. Do you have a preference between the new ENV var or adding a third value to the existing credentials?

AlexCuse:
Not a strong one @cainlevy - I guess maybe including with credentials is slightly more straightforward to document as long as we note the 3rd signing key segment is only respected after v1.X.Y.

cainlevy:
I don't have a strong preference here either. I trust your discretion on it.

AlexCuse:
I took a crack at including with credentials and think I like it better.  Feels like less code, and there was a natural spot for an error in the credential parsing response @cainlevy

PR #241, opened 2024-01-31
oauth: allow passing additional credential data to providers

AlexCuse:


AlexCuse:
This will be needed to pass some additional data (key ID + team ID) to the apple provider.  The signing key could be included in the additional map as well I suppose but since that can be used by all providers I think it makes sense to include as a top-level detail.

AlexCuse:
@cainlevy any concerns about this?  I have the Apple implementation near ready to test built on top of this.

Feeling like it will be time to do configuration files next 😅

cainlevy (approved):
😆 😭 🙈 ok yeah why not

AlexCuse:
Yea might merge with the Apple stuff to keep blame where it belongs for posterity 😅

AlexCuse:
Included in #243

PR #242, opened 2024-02-01
Remove duplicate print of PUBLIC_PORT

cornerman:
First of all. Thank you so much for this project and the great work. For me, it really hits the sweet spot of how to integrate an authentication system :+1: 

Just saw this when starting the server with a public port defined. We already print the port in the Main.


cainlevy (approved):
I see! Yes, good catch. Thanks for the PR. 🙌

PR #243, opened 2024-02-22
oauth: add sign-in with apple support (fixes #110)

### LINKED ISSUE #110: Sign In with Apple support
https://developer.apple.com/sign-in-with-apple/

1. Is this new authentication API easy to integrate into `authn-server`?
2. Is there any work happening on this front?

@cainlevy

AlexCuse:
Implements an oauth provider for sign-in with apple. Requires some
additional flexibility in oauth credentials / providers:

- Include a map of additional data in credentials.
- Allow providers to override secret behavior - configured secret in
apple credentials is a private key used to sign a secret calculated
at runtime, in this case a JWT that includes additional data as claims.
- Allow providers to accept returns as HTTP POST instead of GET.
- Allow providers to add additional oauth options to authorization
request.


lib/oauth/apple/privkey_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
+	"github.com/google/uuid"
+	"github.com/keratin/authn-server/lib/oauth/apple"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// nolint: gosec
```
AlexCuse, line 16: This was not caught locally with golangci-lint v1.53.3 for some reason

AlexCuse:
One thing I noticed so far trying to test is that Apple doesn't like including the scopes in the auth request- requires `response_mode=form_post`.  Hoping omitting scopes until code is exchanged exchange will be a sufficient workaround and won't need to add a post handler for oauth return.

AlexCuse:
> One thing I noticed so far trying to test is that Apple doesn't like including the scopes in the auth request unless `response_mode=form_post` is specified. Should be pretty straightforward but guess we'll need a post handler for oauth return 😢

We just needed to redirect POSTs to the existing handler and allow providers to override the method for routing.

server/public_routes.go
```diff
@@ -87,12 +89,18 @@ func PublicRoutes(app *app.App) []*route.HandledRoute {
 		)
 	}
 
-	for providerName := range app.OauthProviders {
+	for providerName, provider := range app.OauthProviders {
+		var returnRoute *route.Route
+		if provider.ReturnMethod() == http.MethodPost {
```
diegosperes, line 94: For the sake of consistency, I suggest we implement both HTTP methods for all OAuth providers instead of making it conditional.

lib/oauth/apple/idtoken_test.go
```diff
@@ -0,0 +1,295 @@
...
+	"github.com/go-jose/go-jose/v3"
+	"github.com/go-jose/go-jose/v3/jwt"
+	"github.com/google/uuid"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/oauth2"
+)
+
+func testSigningKey(t *testing.T) (jose.SigningKey, string, *rsa.PrivateKey) {
```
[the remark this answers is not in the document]
AlexCuse: Functions in `_test` files aren't callable like that but the change is easy enough to make.

lib/oauth/provider.go
```diff
@@ -21,23 +27,82 @@ type UserInfo struct {
...
+
+type Overrides struct {
+	SecretOverride  SecretOverride
+	AuthCodeOptions AuthCodeOptions
+	MethodOverride  string
+	VerificationKey jose.JSONWebKey
+}
+
 // NewProvider returns a properly configured Provider
 func NewProvider(config *oauth2.Config, userInfo UserInfoFetcher, signingKey jose.SigningKey) *Provider {
```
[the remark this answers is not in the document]
AlexCuse: Good idea!

lib/oauth/apple.go
```diff
@@ -0,0 +1,62 @@
...
+
+	teamID, keyID, expiresIn, constructErr := apple.ExtractCredentialData(credentials.Additional)
+	if constructErr != nil {
+		return nil, fmt.Errorf("apple credentials: %w", constructErr)
+	}
+
+	signingKey, publicKey, constructErr := apple.ParsePrivateKey(credentials.SigningKey, keyID)
+	if constructErr != nil {
+		return nil, fmt.Errorf("failed to parse signing key: %w", constructErr)
+	}
```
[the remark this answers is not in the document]
AlexCuse: Think I prefer leaving it in provider initialization and keeping the app as dumb as possible.

server/public_routes.go
AlexCuse: Think I prefer keeping the surface area as small as possible here.

diegosperes (approved):

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (tr *TokenReader) getAppleIDTokenClaims(t *oauth2.Token) (map[string]interface{}, error) {
+	idTokenVal := t.Extra("id_token")
+	if idTokenVal == nil {
+		return nil, fmt.Errorf("missing id_token")
+	}
+
+	idToken, ok := idTokenVal.(string)
+
+	if !ok {
```
cainlevy, line 45:
```suggestion
	idToken, ok := idTokenVal.(string)
	if !ok {
```

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("id_token is not a string")
+	}
+
+	parsedIDToken, err := jwt.ParseSigned(idToken)
+
+	if err != nil {
```
cainlevy, line 51:
```suggestion
	parsedIDToken, err := jwt.ParseSigned(idToken)
	if err != nil {
```

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, err
+	}
+
+	var hdr *jose.Header
+
+	for i := range parsedIDToken.Headers {
```
cainlevy, line 57:
```suggestion
	var hdr *jose.Header
	for i := range parsedIDToken.Headers {
```

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+		th := parsedIDToken.Headers[i]
+		if th.Algorithm == "RS256" {
+			hdr = &th
+			break
+		}
+	}
+
+	if hdr == nil {
```
cainlevy, line 65:
```suggestion
	}
	if hdr == nil {
```

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("no RS256 key header found")
+	}
+
+	appleRSA, err := tr.keyStore.get(hdr.KeyID)
+
+	if err != nil {
```
cainlevy, line 71:
```suggestion
	appleRSA, err := tr.keyStore.get(hdr.KeyID)
	if err != nil {
```

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("failed to get apple RSA key: %w", err)
+	}
+
+	claims := make(map[string]interface{})
+	err = parsedIDToken.Claims(appleRSA, &claims)
+
+	if err != nil {
```
cainlevy, line 78:
```suggestion
	err = parsedIDToken.Claims(appleRSA, &claims)
	if err != nil {
```

lib/oauth/apple/idtoken.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("failed to verify claims: %w", err)
+	}
+
+	return claims, nil
+}
+
+func extractUserFromClaims(claims map[string]interface{}, clientID string) (string, string, error) {
```
cainlevy, line 85: can we use https://pkg.go.dev/github.com/go-jose/go-jose/v3/jwt#Claims.Validate for the standard `iss`, `aud`, and `exp` claims? then our code can focus on our unique values like `sub` and `email`.

lib/oauth/apple/pubkey.go
```diff
@@ -0,0 +1,136 @@
...
+type mockKeyStore struct {
+	keys map[string]*rsa.PublicKey
+}
+
+func (ks *mockKeyStore) get(keyID string) (*rsa.PublicKey, error) {
+	if key, ok := ks.keys[keyID]; ok {
+		return key, nil
+	}
+	return nil, &keyNotFoundError{keyID: keyID}
+}
```
cainlevy, line 47: should `mockKeyStore` be in `package apple_test`?

lib/oauth/apple/idtoken_test.go
```diff
@@ -0,0 +1,291 @@
+package apple
```
cainlevy, line 1: should this be `package apple_test`?

cainlevy (changes requested):
I can't reasonably verify the Apple-specific details, so I tried to focus on changes outside of `lib/oauth/apple` as well as the JWT validation process.

Would you take a look at the comment on `extractUserFromClaims ` specifically? The rest is aesthetic, I think.

lib/oauth/credentials.go
```diff
@@ -7,19 +7,41 @@ import (
...
-	Secret string
+	ID         string
+	Secret     string
+	Additional map[string]string
 }
 
-// NewCredentials parses a credential string in the format `id:string` and returns a Credentials
-// suitable for OAuth Provider configuration.
+// NewCredentials parses a credential string in the format
+// `id:string:signing_key(optional):additional=data...(optional)`
```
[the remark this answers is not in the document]
AlexCuse: Missed `signing_key` here too will check one last time for remnants of that.

lib/oauth/apple/idtoken_test.go
AlexCuse: Its testing unexported functions that I'm not sure we need to export - but I did move the mock keyStore into the  `_test` file.

Moving to `apple/test` probably works as well.

lib/oauth/apple/idtoken.go
AlexCuse: Will see what I can do here.  The problem is if we read into `jwt.Claims` we lose visibility into `sub` and `email` - but there might be something we can do with a custom `Claims` type in the apple package that embeds / implements its own json unmarshaler that lets us leverage it.

cainlevy: i'd normally verify claims like `iss`, `aud`, and `exp` during the decoding step. is it easier to get at `sub` and `email` if you move the others?

lib/oauth/apple/idtoken_test.go
AlexCuse: Actually this got a lot more practical when I implemented that claims suggestion.

lib/oauth/apple/idtoken.go
AlexCuse: I kinda like how this turned out let me know what you think 95e1df396fb52c19f1f395376896febb5f12d028

lib/oauth/apple/claims.go
```diff
@@ -0,0 +1,40 @@
...
+
+	if c.Expiry == nil {
+		return fmt.Errorf("missing claim 'exp'")
+	}
+
+	if c.IssuedAt == nil {
+		return fmt.Errorf("missing claim 'iat'")
+	}
+
+	// is default 1m leeway OK here?
```
cainlevy, line 35: i'd think so. if there's some network delay when the user returns to authn, this leeway should handle it. if the token is more than 1m too old, then someone's clock is out of sync and they're going to have bigger problems.

lib/oauth/apple/claims.go
```diff
@@ -0,0 +1,40 @@
...
+	return c.Claims.Validate(jwt.Expected{
+		Issuer:   BaseURL,
+		Audience: jwt.Audience{clientID},
+	})
```
cainlevy, line 39: i checked to confirm that it will validate `exp` and `iat` against `time.Now()` by default. 👍

cainlevy (approved):
claims validation looks good to me!

PR #246, opened 2024-03-19
Bugfix: authn initiates user login upon connection attempt

### LINKED ISSUE #245: Authn initiates user login upon connection attempt.
The OAuth endpoint serves three primary functions:

- Creating a new account
- Linking an existing account to a specific provider
- Logging in the user and establishing a valid session

However, there's a problematic behavior that needs to be addressed. Consider this scenario: Account A is already linked to Provider Account Z. Now, if Account B tries to link to the same Provider Account Z, it doesn't return an error as expected. Instead, it logs into Account A.

This behavior is incorrect and poses a significant security risk. It allows Account B unauthorized access to Account A. The correct behavior should be to return an error when Account B attempts to link to Provider Account Z, indicating that the provider account is already linked to another account. This ensures the integrity and security of user accounts.

diegosperes:
Prevent users from gaining unauthorized account access via oauth link.

Fixes #245

### DIFF
--- a/app/services/identity_reconciler.go
+++ b/app/services/identity_reconciler.go
@@ -30,9 +30,14 @@ func IdentityReconciler(accountStore data.AccountStore, cfg *app.Config, provide
 		return nil, errors.Wrap(err, "FindByOauthAccount")
 	}
 	if linkedAccount != nil {
+		if linkedAccount.ID != linkableAccountID && linkableAccountID != 0 {
+			return nil, errors.New("account already linked")
+		}
+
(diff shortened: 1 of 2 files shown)


AlexCuse (approved):

cainlevy (approved):

cainlevy:
Yes, a new version is in progress here: https://github.com/keratin/authn-server/pull/247

AlexCuse:
Just kicked off the 1.18.1 release it should be available shortly @diegosperes @cainlevy 👍

Issue #252, opened 2024-03-26
Oauth: Unlink social account from main authn account


diegosperes:
Authn offers a feature that enables users to connect their accounts with their social media profiles. However, this has created a need for a feature that allows users to disconnect their social media accounts. The current issue is to develop a way in Authn that enables users to unlink social media account.

Proposed Solution:

Create a new HTTP DELETE endpoint to unlink the user's current session from their social media accounts. This endpoint will be public. The frontend needs to adhere to the following contract:

DELETE /oauth/provider

In that case provider can be one of the supported oauth providers like; google

AlexCuse:
This is a good idea.  A few considerations that come to mind - 

- I assume a corresponding public endpoint `GET /oauth/providers` would be needed to enumerate providers enabled for a session and enable the unlink from front end.
- It might make sense to add configured providers to the `GET /accounts/id` endpoint.  A private endpoint to handle unlink by ID for administrative users might be valuable as well.
- do we want to ensure that another means to login to the account is maintained?  Oauth accounts are created with a random password currently.  Password reset is probably an option for most users.
- Is there any revocation we want to support with the provider(s) or is it sufficient to just "forget" the user's oauth account?  I think a user re-entering the oauth flow with a previously issued provider ID shouldn't cause any problems for authn but we'd want to be sure.
- Flip side of above, what if the user revokes access through the provider?  Is there a way to handle re-link in that scenario? (maybe this one is separate issue)

diegosperes:
@AlexCuse Thanks for your comments.

I've submitted a [PR](https://github.com/keratin/authn-server/pull/253) that implements the unlink feature, taking into account all your suggestions during development.

To address your queries:

Regarding whether we need to support any revocation with the provider(s) or if it's sufficient to just "forget" the user's OAuth account, my tests indicate that simply "forgetting" the user's OAuth account allows the user to reenter the OAuth flow without any issues.

As for the scenario where a user revokes access through the provider and how to handle re-linking in that case, this is already implemented. The user simply needs to give consent again to use social login.

PR #253, opened 2024-03-28
Oauth unlink social accounts

diegosperes:
Where the discussion started. [Authn issue.](https://github.com/keratin/authn-server/issues/252)

### Summary
This PR is adding the ability of share oauth account information and delete them.

### Changes

- Adding `email` column to `oauth_accounts` table in order to facilitate users identify their accounts.
- Updating `GET /account/{id}` to return `oauth_providers` containing OAuth information.
- Adding new public endpoint `DELETE /oauth/{provider}` allowing user unlink OAuth accounts.
- Adding new public endpoint `GET /oauth/info` returning oauth information for linked accounts.
- Adding new private endpoint `DELETE /account/{id}/oauth` allowing unlink OAuth accounts for the given user.

There's a particular use case that requires special attention. When a user registers via the OAuth flow, a random password is generated. To unlink the social account, Authn verifies whether the user needs to set a new password. As a side effect, a new error, NEW_PASSWORD_REQUIRED, has been introduced.


kion (approved):

app/data/mysql/migrations.go
```diff
@@ -61,6 +62,18 @@ func createOauthAccounts(db *sqlx.DB) error {
 	return err
 }
 
+func addOauthAccountEmail(db *sqlx.DB) error {
+	_, err := db.Exec(`
+		ALTER TABLE oauth_accounts ADD COLUMN email VARCHAR(255) DEFAULT NULL;
```
AlexCuse, line 67: Storing email like this kind of feels "off" to me, like this is the kind of thing oauth was supposed to prevent almost.  As a consumer of the oauth provider feels like we should be all set once we have a provider ID, and the email is really only needed so we can assign a username.

Is there a use case you had in mind for this beyond including more detail in the list of providers that comes back from the account / info endpoints?  It seems like just the configured provider names should be enough for any I can think of off hand.  I would lean towards simplifying in that way at least for this PR.  Adding columns causes runtime errors for updating users in production if they don't migrate.

If there is a use case you had in mind this could be done as its own PR - this one is already pretty big, and if we make this change I feel like it opens up a need to efficiently update emails when pre-existing oauth users log in and maybe some other stuff I haven't thought of.

diegosperes: @AlexCuse that's a good point! While I understand your concerns about storing email addresses and the potential complexity it adds, there are some compelling reasons to include the user's email in the SQL table:

User Identification: While OAuth provides us with a provider ID, the email address is a more human-readable identifier. It can be useful in administrative tasks, and customer support.

Communication: Having the user's email address readily available in our database allows us to easily communicate with our users.

User Experience: If a user forgets which OAuth provider they used to sign up, having their email stored can help us guide them to the correct provider.

docs/api.md
```diff
@@ -658,6 +697,62 @@ If the OAuth process failed, the redirect will have `status=failed` appended to
...
+
+Visibility: Public
+
+`DELETE /oauth/:providerName`
+
+| Params | Type | Notes |
+| ------ | ---- | ----- |
+| `providerName` | string | * google |
+
+Delete an OAuth account from the current session. If the session was initiated via the OAuth flow, the account will be flagged to reset the password during the next sign-in attempt.
```
AlexCuse, line 736: Is this language right?  It looked to me like the removal would be blocked if a password change was required but maybe I misread.  I hope we wouldn't reset the password the next time someone tried to login based on just the emaiil.

server/handlers/delete_account_oauth_test.go
```diff
@@ -0,0 +1,66 @@
...
+	server := test.Server(app)
+	defer server.Close()
+
+	client := route.NewClient(server.URL).Authenticated(app.Config.AuthUsername, app.Config.AuthPassword)
+
+	http.DefaultClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		return http.ErrUseLastResponse
+	}
+
+	t.Run("delete social account", func(t *testing.T) {
```
AlexCuse, line 26: This test could probably just be called "success"

server/handlers/delete_account_oauth_test.go
```diff
@@ -0,0 +1,66 @@
...
+		account, err := app.AccountStore.Create("deleted-social-account@keratin.tech", []byte("password"))
+		require.NoError(t, err)
+
+		time.Sleep(5 * time.Second)
```
AlexCuse, line 30: This is a long sleep for a test what is it needed for?

server/handlers/delete_oauth_test.go
```diff
@@ -0,0 +1,78 @@
...
+	client := route.NewClient(server.URL).Referred(&app.Config.ApplicationDomains[0]).WithCookie(&http.Cookie{
+		Name:  app.Config.OAuthCookieName,
+		Value: "",
+	})
+
+	http.DefaultClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		return http.ErrUseLastResponse
+	}
+
+	t.Run("unanthorized", func(t *testing.T) {
```
AlexCuse, line 35: unauthorized

server/handlers/delete_account_oauth_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+		payload := map[string]interface{}{"oauth_providers": []string{"test"}}
+		res, err := client.DeleteJSON("/accounts/1/oauth", payload)
+		require.NoError(t, err)
+
+		require.Equal(t, http.StatusOK, res.StatusCode)
+		require.Equal(t, []byte{}, test.ReadBody(res))
+	})
+
+	t.Run("return not found when user does not exists", func(t *testing.T) {
```
AlexCuse, line 43: "user does not exist" - generally shorter test names will read better in output on failures.  And the assertions should take care of the "return not found" stuff for you.  It is frustrating when the assertions change but the test names don't.

server/handlers/delete_oauth_test.go
```diff
@@ -0,0 +1,78 @@
...
+		res, err := client.Delete("/oauth/test")
+		require.NoError(t, err)
+
+		require.Equal(t, http.StatusUnauthorized, res.StatusCode)
+		require.Equal(t, []byte{}, test.ReadBody(res))
+	})
+
+	t.Run("delete social account", func(t *testing.T) {
```
AlexCuse, line 43: "success"

server/handlers/delete_oauth_test.go
```diff
@@ -0,0 +1,78 @@
...
+		session := test.CreateSession(app.RefreshTokenStore, app.Config, account.ID)
+
+		res, err := client.WithCookie(session).Delete("/oauth/test")
+		require.NoError(t, err)
+
+		require.Equal(t, http.StatusOK, res.StatusCode)
+		require.Equal(t, []byte{}, test.ReadBody(res))
+	})
+
+	t.Run("return unprocessable entity when user requires a new password", func(t *testing.T) {
```
AlexCuse, line 61: "requires new password"

server/handlers/get_account_test.go
```diff
@@ -35,32 +36,62 @@ func TestGetAccount(t *testing.T) {
...
-	assert.Equal(t, false, responseData.Deleted)
+	sort.Slice(oAccs, func(i, j int) bool {
+		return oAccs[i].Provider < oAccs[j].Provider
+	})
+
+	assert.Equal(t, []string{"application/json"}, res.Header["Content-Type"])
+	assert.Equal(t, responseData, response{
+		ID:       acc.ID,
+		Username: acc.Username,
+		OauthAccounts: []map[string]interface{}{
```
AlexCuse, line 80: Its weird to expect two here, same with the hard coded providers.  You could just loop over oAccs and build up the map I think?

server/handlers/get_account_test.go
```diff
@@ -35,32 +36,62 @@ func TestGetAccount(t *testing.T) {
...
+				"email":               oAccs[0].Email,
+			},
+			{
+				"provider":            "trial",
+				"provider_account_id": oAccs[1].ProviderID,
+				"email":               oAccs[1].Email,
+			},
+		},
+		LastLoginAt:       "",
+		PasswordChangedAt: acc.PasswordChangedAt.Format("2006-01-02T15:04:05Z07:00"),
```
AlexCuse, line 93: Can use `time.RFC3339` here.

server/handlers/get_oauth_info_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+	server := test.Server(app)
+	defer server.Close()
+
+	client := route.NewClient(server.URL).Referred(&app.Config.ApplicationDomains[0]).WithCookie(&http.Cookie{
+		Name:  app.Config.OAuthCookieName,
+		Value: "",
+	})
+
+	t.Run("unanthorized", func(t *testing.T) {
```
AlexCuse, line 23: "unauthorized"

server/handlers/get_oauth_info_test.go
```diff
@@ -0,0 +1,47 @@
...
+		res, err := client.Get("/oauth/info")
+		require.NoError(t, err)
+
+		require.Equal(t, http.StatusUnauthorized, res.StatusCode)
+		require.Equal(t, []byte{}, test.ReadBody(res))
+	})
+
+	t.Run("get oauth info", func(t *testing.T) {
```
AlexCuse, line 31: "success"

server/public_routes.go
```diff
@@ -54,6 +54,10 @@ func PublicRoutes(app *app.App) []*route.HandledRoute {
 		route.Delete("/totp").
 			SecuredWith(originSecurity).
 			Handle(handlers.DeleteTOTP(app)),
+
+		route.Get("/oauth/info").
```
AlexCuse, line 58: Maybe a better route name would be `/oauth/accounts` or even just `/oauth`

docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
       - "${REDIS_PORT:-8701}:6379"
 
   mysql:
-    platform: linux/x86_64
+    # platform: linux/x86_64
```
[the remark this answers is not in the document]
diegosperes: I'm not entirely certain, but commenting out this line seems to make it work on my local setup. This change was introduced in a specific [commit](https://github.com/keratin/authn-server/commit/bbc6311ae913219a51f8fcf15176cb4281994a7b#diff-e45e45baeda1c1e73482975a664062aa56f20c03dd9d64a827aba57775bed0d3R9). What are your thoughts on removing this line?

docs/api.md
```diff
@@ -295,6 +306,34 @@ Visibility: Private
...
+### Delete OAuth account by user id
+
+Visibility: Private
+
+`DELETE /accounts/:id/oauth`
+
+|        Params       |   Type  |      Notes      |
+| ------------------- | ------- | --------------- |
+| `id`                | integer | User account Id |
+| `oauth_providers`   | integer | User account Id |
```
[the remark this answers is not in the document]
diegosperes: Good catch! The doc is wrong here, is suppose to be a provider name.

server/handlers/delete_account_oauth.go
```diff
@@ -0,0 +1,59 @@
...
+	"github.com/gorilla/mux"
+	"github.com/keratin/authn-server/app"
+	"github.com/keratin/authn-server/app/services"
+	"github.com/keratin/authn-server/lib/parse"
+)
+
+func DeleteAccountOauth(app *app.App) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		isNotFoundErr := func(err error) (string, bool) {
+			fieldErr, ok := err.(services.FieldErrors)
```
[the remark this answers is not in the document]
diegosperes: Thanks!

docs/api.md
diegosperes: Have improved the doc here

docs/api.md
```diff
@@ -658,6 +697,62 @@ If the OAuth process failed, the redirect will have `status=failed` appended to
...
+
+#### Delete OAuth account
+
+Visibility: Public
+
+`DELETE /oauth/:providerName`
+
+| Params | Type | Notes |
+| ------ | ---- | ----- |
+| `providerName` | string | * google |
```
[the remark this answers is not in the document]
diegosperes: doc updated

server/handlers/delete_account_oauth_test.go
diegosperes: to avoid returns an errors requesting to reset password

server/handlers/get_account_test.go
diegosperes: Don't think it's a good idea, because if an OAuth provider is disabled after a few registrations won't ever be returned again.

server/public_routes.go
diegosperes: I like the idea of info here, make easier to extend datapoints.

app/data/mysql/migrations.go
diegosperes: @AlexCuse I have implemented you suggestion 😄

app/services/account_oauth_ender.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	mappedProviders := map[string]uint8{}
+	for _, provider := range providers {
+		mappedProviders[strings.ToLower(provider)] = 1
+	}
+
+	for _, oAccount := range oauthAccounts {
+		_, isProviderMatched := mappedProviders[strings.ToLower(oAccount.Provider)]
+		hasRandomOauthPassword := math.Abs(oAccount.CreatedAt.Sub(account.PasswordChangedAt).Seconds()) < 5
```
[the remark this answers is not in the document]
diegosperes: I'm not entirely sure I understand the context, but in this scenario, due to the use of `math.Abs`, the modulus of the date will exceed 5 seconds, thus not triggering a password reset request.

docker-compose.yml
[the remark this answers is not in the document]
cainlevy: Removing the line seems fine to me

app/services/account_oauth_ender.go
```diff
@@ -0,0 +1,47 @@
...
+
+		if hasRandomOauthPassword && isProviderMatched {
+			return FieldErrors{{"password", ErrNewPasswordRequired}}
+		}
```
cainlevy, line 36: I don't understand why this is helpful. Why should an account be required to set a password _before_ disconnecting an OAuth identity? Why can't it set the password afterwards?

app/services/account_oauth_ender.go
```diff
@@ -0,0 +1,47 @@
+package services
+
+import (
+	"math"
+	"strings"
+
+	"github.com/keratin/authn-server/app/data"
+)
+
+func AccountOauthEnder(store data.AccountStore, accountId int, providers []string) error {
```
cainlevy, line 10: I would suggest `IdentityRemover`.

It's quite likely that all of the "Identity" service functions should have a different noun, but I would request handling that as a separate branch.

app/services/account_oauth_getter.go
```diff
@@ -0,0 +1,28 @@
+package services
+
+import (
+	"github.com/keratin/authn-server/app/data"
+	"github.com/pkg/errors"
+)
+
+func AccountOauthGetter(accountStore data.AccountStore, accountID int) ([]map[string]interface{}, error) {
```
cainlevy, line 8: Since this data is returned as part of the account JSON, you might consider implementing it within the existing `AccountGetter`.

server/handlers/delete_account_oauth.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/gorilla/mux"
+	"github.com/keratin/authn-server/app"
+	"github.com/keratin/authn-server/app/services"
+	"github.com/keratin/authn-server/lib/parse"
+)
+
+func DeleteAccountOauth(app *app.App) http.HandlerFunc {
```
cainlevy, line 14: it looks like this endpoint is built to delete an arbitrary number of providers at once. this seems more complicated than necessary (the JSON body parsing), and perhaps also not likely to be useful with a typical CRUD-style user interface.

which do you think would be most useful:

1. `DELETE /accounts/:id/oauth` - deletes all oauth records
2. `DELETE /accounts/:id/oauth/:name` - deletes one oauth record

server/handlers/get_account.go
```diff
@@ -28,6 +28,12 @@ func GetAccount(app *app.App) http.HandlerFunc {
 			panic(err)
 		}
 
+		oautAccounts, err := services.AccountOauthGetter(app.AccountStore, id)
```
cainlevy, line 31:
```suggestion
		oauthAccounts, err := services.AccountOauthGetter(app.AccountStore, id)
```

(but also, you might consider handling this in `services.AccountGetter`)

server/public_routes.go
cainlevy: `/oauth/info` sounds like an endpoint that will tell me about the server's oauth configuration.

If we want to introduce "tell me about my own account" endpoints, then here are a couple ideas:

* `/account`
* `/me`

Both might return the same payload as `/account/:id`.

cainlevy (changes requested):
Thanks for working on this! I'm sorry for the delay in reviewing.

app/services/account_oauth_getter.go
diegosperes: Nice one! I have updated the code accordingly.

server/handlers/delete_account_oauth.go
diegosperes: Code updated to use 2. I don't have a strong preference, but the second approach indeed aligns more closely with the CRUD-style.

server/handlers/get_account.go
diegosperes: I have moved the logic to `services.AccountGetter` 😄

server/public_routes.go
diegosperes: Thanks @AlexCuse and @cainlevy for the heads up, I have changed the implementation to use `/oauth/accounts`

app/services/account_oauth_ender.go
diegosperes: @cainlevy agree - I can certainly rename these functions in a separate PR. For now, it would be more beneficial to limit the scope of changes to avoid increasing the size of the current PR.

app/services/account_getter.go
```diff
@@ -1,11 +1,47 @@
...
 
 import (
+	"time"
+
 	"github.com/keratin/authn-server/app/data"
 	"github.com/keratin/authn-server/app/models"
 	"github.com/pkg/errors"
 )
 
+func AccountToJson(account *models.Account) map[string]interface{} {
```
AlexCuse, line 11: Don't think this makes a ton of sense in services.  Maybe implementing `json.Marshaler` on the account and oauth account models would make sense?

https://pkg.go.dev/encoding/json#Marshaler

server/handlers/get_account_test.go
AlexCuse: I don't understand - the oauth accounts passed in already have a provider that is used for sorting a few lines up. Shouldn't that same provider be used to populate the maps?

Don't understand how disabled providers come into play here either.

diegosperes: I misread the code, code updated 😄 .

app/services/account_oauth_ender.go
diegosperes: Thanks for share your thoughts, I have removed this validation from the code.

app/services/account_oauth_ender.go
diegosperes: Thanks for the suggestions, the function was renamed to `IdentityRemover`.

app/services/account_getter.go
diegosperes: Nice one! Account and oauth account now implements Marshaler interface.

diegosperes:
@AlexCuse @cainlevy Thanks very much from your comments, I have updated the code accordingly to them 😄

app/models/account.go
```diff
@@ -30,3 +32,25 @@ func (a Account) TOTPEnabled() bool {
...
+		formattedLastLogin = a.LastLoginAt.Format(time.RFC3339)
+	}
+
+	formattedPasswordChangedAt := ""
+	if !a.PasswordChangedAt.IsZero() {
+		formattedPasswordChangedAt = a.PasswordChangedAt.Format(time.RFC3339)
+	}
+
+	return json.Marshal(map[string]interface{}{
+		"id":                  a.ID,
```
AlexCuse, line 48: inline structs with JSON tags can be helpful here but I think this is fine.

app/services/account_getter_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"sort"
+	"testing"
+
+	"github.com/keratin/authn-server/app/data/mock"
+	"github.com/keratin/authn-server/app/services"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAccountGetter(t *testing.T) {
+	t.Run("returns empty map when no oauth accounts", func(t *testing.T) {
```
AlexCuse, line 13: Hopefully not something we'll see at runtime but its probably worth adding a test for not found and the case where the oauth account lookup errors while we are here.

app/services/account_oauth_ender.go
```diff
@@ -0,0 +1,25 @@
+package services
+
+import (
+	"github.com/keratin/authn-server/app/data"
+)
+
+func IdentityRemover(store data.AccountStore, accountId int, providers []string) error {
```
AlexCuse, line 7: Rename file to `identity_remover.go`

app/services/account_oauth_ender_test.go
```diff
@@ -0,0 +1,62 @@
+package services_test
```
AlexCuse, line 1: rename file

app/services/identity_reconciler.go
```diff
@@ -68,11 +73,37 @@ func IdentityReconciler(accountStore data.AccountStore, cfg *app.Config, provide
...
+
+	for _, oAccount := range oAccounts {
+		if providerName != oAccount.Provider && providerUser.ID != oAccount.ProviderID {
+			continue
+		}
+
+		if oAccount.Email == "" {
+			_, err = accountStore.UpdateOauthAccount(accountID, oAccount.Provider, providerUser.Email)
+			if err != nil {
+				return errors.Wrap(err, "UpdateOauthAccountEmail")
```
AlexCuse, line 103: Wrapping annotation here should be UpdateOauthAccount

app/services/identity_reconciler.go
```diff
@@ -38,12 +38,17 @@ func IdentityReconciler(accountStore data.AccountStore, cfg *app.Config, provide
 			return nil, errors.New("account locked")
 		}
 
+		err = updateUserInfo(accountStore, linkedAccount.ID, providerName, providerUser)
```
AlexCuse, line 41: Should `linkedAccount` have the associated oauth accounts included so it can be passed directly and skip the oauth account lookup in this function?

app/services/identity_reconciler.go
```diff
@@ -68,11 +73,37 @@ func IdentityReconciler(accountStore data.AccountStore, cfg *app.Config, provide
...
+	if len(oAccounts) == 0 {
+		return nil
+	}
+
+	for _, oAccount := range oAccounts {
+		if providerName != oAccount.Provider && providerUser.ID != oAccount.ProviderID {
+			continue
+		}
+
+		if oAccount.Email == "" {
```
AlexCuse, line 100: Is email changing a valid use case here?  Eg switching between apple private relay and "real" email address, or maybe changing email associated with a facebook account?  It may be easy enough to cover/test that we should include it even if unlikely.

app/services/identity_reconciler_test.go
```diff
@@ -83,4 +83,25 @@ func TestIdentityReconciler(t *testing.T) {
 		assert.Error(t, err)
 		assert.Nil(t, found)
 	})
+
+	t.Run("update missing email after oauth migration table", func(t *testing.T) {
```
AlexCuse, line 87: Test "changed" email if support is added.

AlexCuse (changes requested):
This is looking really good think this is my last round of comments.

AlexCuse:
@cainlevy do you think we need to worry about revoking the stored access token here or is it sufficient to "forget"?  Looking at apple's implementation it would be fairly simple, though it does require an additional URL to be defined https://developer.apple.com/documentation/sign_in_with_apple/revoke_tokens/

A quick search on google and microsoft seems to indicate the tokens expire after an hour anyway so may be nothing to worry about there.

app/services/identity_reconciler.go
diegosperes: Completely agree with you. In this situation, I think it would be more beneficial to handle the existing technical debt separately. This is because addressing it will require a significant amount of changes everywhere.

app/services/identity_reconciler.go
diegosperes: I'm not entirely certain about these use cases. Given that the queries take into account the provider ID, do you think it's significant?

AlexCuse: I'm not sure either but if the goal of capturing email is for end user communcation I think you want it to be right / up to date.

app/services/identity_reconciler_test.go
```diff
@@ -84,7 +84,7 @@ func TestIdentityReconciler(t *testing.T) {
 		assert.Nil(t, found)
 	})
 
-	t.Run("update missing email after oauth migration table", func(t *testing.T) {
+	t.Run("update changed email after oauth migration table", func(t *testing.T) {
```
AlexCuse, line 87: This is still testing "update missing email" - I meant should add a test if you added support for the email change discussed above.

server/handlers/get_oauth_accounts_test.go
```diff
@@ -42,6 +50,12 @@ func TestGetOauthInfo(t *testing.T) {
 		require.NoError(t, err)
 
 		require.Equal(t, http.StatusOK, res.StatusCode)
-		require.Equal(t, []byte(expected), test.ReadBody(res))
+
+		json.Unmarshal(test.ReadBody(res), &expected)
```
AlexCuse, line 54: Lint error here (need to assert non-nil error)

docs/api.md
```diff
@@ -295,6 +306,26 @@ Visibility: Private
       ]
     }
 
+### Delete OAuth account by user id
```
cainlevy, line 309: these docs may need a quick update to match the changes in the `delete_account_oauth` handler

cainlevy (approved):
thank you! it's looking good.

cainlevy:
> do you think we need to worry about revoking the stored access token here or is it sufficient to "forget"?

Personally I would expect it's enough to forget them. Revoking them sounds very thorough but also would require a higher level of integration with each identity provider.

docs/api.md
diegosperes: Thanks @cainlevy, doc updated 😄

AlexCuse (approved):

PR #256, opened 2024-04-05
Fix links in readme

cornerman:
Just stumbled over the broken links and missing image in the readme.


cainlevy (approved):
Thanks for catching those

PR #259, opened 2024-04-17
fix: null OAuth email behavior

diegosperes:
### Summary

The problem arises when Authn creates a new email column and sets its default value to NULL. This causes an error, impacting all users who previously used their social login. 

### Changes

- Refactoring code to accommodate NULL values in the OAuth email column, as the default database value is NULL

### DIFF
--- a/app/data/mysql/account_store_test.go
+++ b/app/data/mysql/account_store_test.go
@@ -1,8 +1,10 @@
 package mysql_test
 
 import (
+	"database/sql"
 	"testing"
 
+	"github.com/keratin/authn-server/app/data"
 	"github.com/keratin/authn-server/app/data/mysql"
 	"github.com/keratin/authn-server/app/data/testers"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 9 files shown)


app/data/mysql/account_store_test.go
```diff
@@ -11,10 +13,37 @@ import (
...
-	store := &mysql.AccountStore{db}
+	var store data.AccountStore = &mysql.AccountStore{db}
 	for _, tester := range testers.AccountStoreTesters {
 		db.MustExec("TRUNCATE accounts")
 		db.MustExec("TRUNCATE oauth_accounts")
 		tester(t, store)
 	}
+
+	t.Run("handle oauth email with null value", func(t *testing.T) {
+		db := store.(interface {
```
AlexCuse, line 24: Why do you need to redefine `db` here?  Seems like the one defined in l14 should work?

app/data/mysql/account_store_test.go
```diff
@@ -11,10 +13,37 @@ import (
 func TestAccountStore(t *testing.T) {
 	db, err := mysql.TestDB()
 	require.NoError(t, err)
-	store := &mysql.AccountStore{db}
+	var store data.AccountStore = &mysql.AccountStore{db}
```
AlexCuse, line 16: Not sure I understand the need for declaration change here.

app/data/postgres/account_store_test.go
```diff
@@ -38,10 +40,37 @@ func newTestDB() (*sqlx.DB, error) {
 func TestAccountStore(t *testing.T) {
 	db, err := newTestDB()
 	require.NoError(t, err)
-	store := &postgres.AccountStore{db}
+	var store data.AccountStore = &postgres.AccountStore{db}
```
AlexCuse, line 43: Same questions as above for this file.

diegosperes: Same answer above!

app/data/mysql/account_store_test.go
diegosperes: Due to the issue below, previously store wasn't an interface.

Screenshot 2024-04-17 at 19 54 06

app/data/mysql/account_store_test.go
diegosperes: It's because `data.AccountStore` does not implements `Exec` but `sqlx.Ext` does, we can replace that hardcoded interface for `sqlx.Ext` to make it clear. What do you think?

diegosperes: @AlexCuse Good catch! Thanks for that

app/data/mysql/account_store_test.go
diegosperes: I have reverted that change now

AlexCuse (approved):

Issue #163, opened 2020-11-26
Does authn-server provide Twitter OAuth?


lkeix:
Hi!
Does authn-server provide Twitter OAuth?
If it dosen't, do you have plan to be compatible with Twitter OAuth?

cainlevy:
Hello! Twitter has not been implemented yet. If you or anyone else reading this thread is interested in contributing a pull request, the requirements are roughly:

1. add a new ENV var parser (e.g. `TWITTER_OAUTH_CREDENTIALS`)
2. implement a function that uses the final token to retrieve the user's ID & email
3. add to documentation

Here's a recent example for Microsoft OAuth: https://github.com/keratin/authn-server/commit/f6ea426503709e4456b09f7a4bee37b7a616798d

lkeix:
Ok!
I grasp roughly requirements. I'll implement the features, send PR.

lkeix:
Hi!
I'm implementing Twitter oauth now(I added Twitter Authorization ENV). Twitter OAuth seem more complex than other service's oauth. Twitter's oauth use 3-legged authorization flow. (Please look at detail -> https://developer.twitter.com/ja/docs/basics/authentication/overview/3-legged-oauth)
While, Twitter provide oauth2.0. But, I couldn't get access token through oauth2.0. Twitter oauth2.0 seems can access public data only.

So, we have no choice to use oauth1.0 when we get authorization token from twitter.
Can I implement twitter oauth feature by using other oauth1.0 library(e.g. https://github.com/gomodule/oauth1)?

cainlevy:
Sorry, I'm not clear on the issue. AuthN implements most of the 3-legged OAuth2.0 flow. It only requires a custom function to find basic user information at the end, after the request token has been exchanged for the access token.

One complication I've found is that Twitter apps must be configured with permission to ask for email addresses:

> The "Request email addresses from users" checkbox is available under the app permissions on developer.twitter.com. Privacy Policy URL and Terms of Service URL fields must be completed in the app settings in order for email address access to function. If enabled, users will be informed via the oauth/authorize dialog that your app can access their email address.

https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials

fdietze:
@lkeix do you have any incomplete code to continue? I might give it a shot.

PR #263, opened 2024-08-09
twitter oauth

fdietze:
I'm still trying to get this working locally...

closing #163

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -132,6 +132,9 @@ func initializeOAuthProviders(cfg *Config) (map[string]oauth.Provider, error) {
 	if cfg.FacebookOauthCredentials != nil {
 		oauthProviders["facebook"] = *oauth.NewFacebookProvider(cfg.FacebookOauthCredentials)
 	}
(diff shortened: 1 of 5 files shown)


Sassoo75543 (commented):
Good

fdietze:
I'm not able to get past the first step. After getting forwarded in the browser to twitter, it always tells me that `Something went wrong`. But I don't know what the error could be. I tried several different Oauth scopes, also no scopes at all. I ran out of ideas. Any suggestions?

AlexCuse:
This seems like one of those "lost to time" comments but have you tried base64 encoding the client ID and secret @fdietze https://stackoverflow.com/a/77712773

Ran into a lot of weird undocumented stuff working on apple happy to try and help with this if I can.

fdietze:
@AlexCuse Thanks! I'll probably get back to it next week, I'll report here.

fdietze:
As you already noticed, I didn't get back to it, because priorities changed. I hope this can be a good starting point for future work.

Issue #264, opened 2025-01-01
Account by username


ahme-dev:
Is it possible to get an account using the username only, rather than the id?

cainlevy:
No, currently the GET endpoint only finds by ID

ahme-dev:
> No, currently the GET endpoint only finds by ID

Is it possible for this to be added as a feature? I could try making a PR for it if I find the time. 

At the moment, when we create an auth account, we run a query to store the account id and username on our end, to get the account with later. Somehow the latter query fails sometimes, which essentially locks us out of getting the auth account, since we can't link the username to the account id, and therefore can't get the auth account. If I'm missing some feature or way, I'd appreciate a correction. (we've set the signup endpoint to private so it only functions on the backend)

cainlevy:
Yes, I think this is possible and would be happy to look at a PR. Perhaps the existing GET endpoint could search both fields in sequence?

It sounds like you're dealing with a two-phase commit problem where your client POSTS an AuthN account but then fails to POST that information to your application. If the user tries to sign up again, they'll get an "already taken" error from AuthN. Is that right?

This can be challenging. One approach you might consider is turning the sign-up process into two steps, where the first step is a find-or-create action to AuthN and the second step collects any additional information you need to save a user profile in your application. This is similar to an OAuth flow, where authentication happens elsewhere and your application must have a separate process for gathering profile information.

PR #265, opened 2025-01-27
Add username to get account

ahme-dev:
Adds the ability to get an account by username, alongside the normal id. As discussed in #264

### DIFF
--- a/app/services/account_getter.go
+++ b/app/services/account_getter.go
@@ -6,16 +6,37 @@ import (
 	"github.com/pkg/errors"
 )
 
-func AccountGetter(store data.AccountStore, accountID int) (*models.Account, error) {
-	account, err := store.Find(accountID)
-	if err != nil {
-		return nil, errors.Wrap(err, "Find")
+type AccountGetterParams struct {
+	AccountID *int
+	Username  *string
+}
+
+func AccountGetter(store data.AccountStore, params AccountGetterParams) (*models.Account, error) {
+	var account *models.Account
+
+	if params.AccountID != nil {
+		ac, err := store.Find(*params.AccountID)
+		if err != nil {
+			return nil, errors.Wrap(err, "Find")
+		}
+
+		account = ac
 	}
+
+	if params.Username != nil && params.AccountID == nil {
+		ac, err := store.FindByUsername(*params.Username)
+		if err != nil {
+			return nil, errors.Wrap(err, "FindByUsername")
+		}
+
+		account = ac
+	}
+
 	if account == nil {
 		return nil, FieldErrors{{"account", ErrNotFound}}
 	}
 
-	oauthAccounts, err := store.GetOauthAccounts(accountID)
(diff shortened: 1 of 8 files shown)


app/services/account_getter.go
```diff
@@ -6,16 +6,37 @@ import (
...
-func AccountGetter(store data.AccountStore, accountID int) (*models.Account, error) {
-	account, err := store.Find(accountID)
-	if err != nil {
-		return nil, errors.Wrap(err, "Find")
+type AccountGetterParams struct {
+	AccountID *int
+	Username  *string
+}
+
+func AccountGetter(store data.AccountStore, params AccountGetterParams) (*models.Account, error) {
```
ahme-dev, line 14: Haven't seen structs used for params, so let me know if this should be changed

server/handlers/get_account.go
```diff
@@ -11,16 +11,27 @@ import (
...
+			paramID = &id
+		}
+
+		account, err := services.AccountGetter(app.AccountStore, services.AccountGetterParams{
+			AccountID: paramID,
+			Username:  paramUsername,
+		})
 		if err != nil {
 			if _, ok := err.(services.FieldErrors); ok {
-				WriteNotFound(w, "account")
```
cainlevy, line 23: Can you say more about this change? I'm not sure how it's related.

cainlevy (changes requested):
Hi @ahme-dev, thanks for your patience. This looks good! I have one small question before I'm ready to approve.

server/handlers/get_account.go
ahme-dev: I've commited it by mistake, re-adding now. Let me know if there's anything else

cainlevy (approved):
Thanks @ahme-dev

cainlevy:
@ahme-dev Would you take a look at the failed tests in https://github.com/keratin/authn-server/actions/runs/13155722367/job/36867948456?pr=265?

I believe the failed lint is a different issue.

server/private_routes.go
```diff
@@ -32,7 +32,7 @@ func PrivateRoutes(app *app.App) []*route.HandledRoute {
 			SecuredWith(authentication).
 			Handle(handlers.PostAccountsImport(app)),
 
-		route.Get("/accounts/{id:[0-9]+}").
+		route.Get("/accounts/{id}").
```
ahme-dev, line 35: @cainlevy, one of the tests (for available account) fails because this overrides the public route "/accounts/available". Any suggestions to work around this? Can't use the older [0-9] since it rejects accept emails. I think we ideally want to put the overriden route above this route definition, but being in separate files, it might be difficult.

cainlevy: Oh yes that is awkward.

I think it's worth trying to register public routes before private routes to fix this. It sounds like the test case, as written, will ensure it remains that way. If we later run into a new issue that implies a need to reverse the order again, we'll find a new kind of solution.

